PR #17, opened 2020-02-05
Change shebang to use python3

nicomem:
The command `/usr/bin/env python` uses the Python 2 interpreter (at least on my machine).

By changing the shebang to explicitly use the Python3 interpreter, the problem should not happen anymore.


master-of-zen:
Thanks :)

PR #19, opened 2020-02-06
Fix manually specifying scenedetect threshold

n9Mtq4:
The threshold being passed to scenedetect needs be a float, but is set as a string in the option parser. This leads the scene detection failing with `Error in PySceneDetect` when setting `-tr 40`. This PR sets the type of the threshold option to be a float.


master-of-zen:
Thanks :)

PR #20, opened 2020-02-06
Replace os.system with subprocess.call and use it for logging

nicomem:
The python docs recommends the use of the `subprocess` module over the `os.system` function.

It also makes logging easier and cross-platform compared to the existing solution (and also solves a problem I had where ffmpeg didn't create the segment files)


master-of-zen:
Good work, Thanks. That help a lot :)

PR #21, opened 2020-02-07
Fix and simplify audio extraction

nicomem:
This PR fixes the audio extraction and merging (the ouput video didn't contain the audio part on my machine).

It also simplifies the process by:
- not creating the `audio_check.txt` temp file and capturing the ffprobe output in a variable
- defaulting the audio params to `-c:a copy` thus allowing code merging


master-of-zen:
Thank you, will check is everything works properly and merge :)

master-of-zen:
File with no audio gives concatenation error:
`UnboundLocalError: local variable 'audio' referenced before assignment`

PR #27, opened 2020-02-12
Reuse past work if same input file

nicomem:
This PR try to reuse work that has already been done in an interrupted past work if the input file is the same.

## How it works & main points:
- Hash the input file and compare it to the saved input file hash (`.temp/input.hash`), if they are the same, we can reuse the work (else, we throw the directory like before)
- It reuses:
  - The split input files
  - The split audio file
  - The encoded files
- Because ffmpeg creates the output file before it is finished, another directory has been added which holds the encoded files (`.temp/done`)

## Other points:
- The hash function used is blake2, a fast hash function. Its successor [blake3 is even faster](https://github.com/BLAKE3-team/BLAKE3/) but there is no easy way of using it at the moment in Python (except incorporating the `b3sum` binary in the repo)
- Since the split input files are reused, the `--scenes` option doesn't have much use other than deactivating the splitting (maybe it can be renamed and only do that)
- Since the last element in the commands was not used and differed between the encoders, it was modified to hold the file name (which was needed to move the output to the `done` directory)

### DIFF
--- a/av1an.py
+++ b/av1an.py
@@ -4,35 +4,39 @@
 # Benchmarking
 # Add conf file
 
-from tqdm import tqdm
 import sys
 import os
 import shutil
 from os.path import join
-from psutil import virtual_memory
 import argparse
 from math import ceil
 from multiprocessing import Pool
 import multiprocessing
 import subprocess
 from pathlib import Path
-from typing import Optional, Union
+from typing import Optional, Union, Tuple, List
+from hashlib import blake2b
 
-try:
-    import scenedetect
-except ImportError:
-    print('PySceneDetect not found. Please check installation')
-    sys.exit()
-
-
-from scenedetect.video_manager import VideoManager
-from scenedetect.scene_manager import SceneManager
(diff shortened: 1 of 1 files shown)


master-of-zen:
A lot of work :) 
But i use another way that is more simple and straight. Adding `--resume`
In branch  [Encoding-resume](https://github.com/master-of-zen/Av1an/tree/Encoding-resume). If `--resume` and `.temp` exist just skipping splitting all together. 
Later will be added feature that finished encodes are tracked, so you can resume from any point of encoding without loosing finished progress.
This will not be merged, but i will include all small changes and improvements :1st_place_medal:  cheers

master-of-zen:
+  the problem is that you can't be sure is file encoded properly or not, so additional check is required

master-of-zen:
I will reuse most of your pull request, but need to figure out what to do with encoded chunks

master-of-zen:
@nicomem Thank you for your work :+1: i included all improvements :)

PR #40, opened 2020-03-24
Customize .temp folder with --temp_dir

natis1:
This will, for example, allow you to run multiple instances of av1an from the same working path.

Also removes all explicit references to ".temp"


master-of-zen:
Done #https://github.com/master-of-zen/Av1an/commit/8b89405d152ad24439fc6aee09b59a40f6d39e56

PR #43, opened 2020-04-17
Fix python version requirement

6ws:


master-of-zen:
Fixed in https://github.com/master-of-zen/Av1an/commit/c0142e71a246b94d81ef68fe7eefeed6e81b47ad

PR #46, opened 2020-04-19
Asks user if they would like to delete their temp folder.

### LINKED ISSUE #45: Error if .temp/split already exists but is empty.
If the directory .temp/split exists but is empty it errors out.

Maybe it should offer the user the option to delete the existing .temp folder?

nathanielcwm:
Fixes #45 

~~I used the walrus operator so it now requires python 3.8. I would be happy to change it though.~~

### DIFF
--- a/av1an.py
+++ b/av1an.py
@@ -7,6 +7,7 @@ import sys
 import os
 import shutil
 from distutils.spawn import find_executable
+from distutils.util import strtobool
 from ast import literal_eval
 from psutil import virtual_memory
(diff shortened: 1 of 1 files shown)


master-of-zen:
Different solution https://github.com/master-of-zen/Av1an/commit/90855f61165db0ddd03c972b03cea92c9e05eb4b

master-of-zen:
Thanks for effort :1st_place_medal:

PR #50, opened 2020-04-24
Restore terminal settings before exit

### LINKED ISSUE #49: Leaves commandline in unusable state
Running on Ubuntu 18.04, python 3.8.2. After running the main script, bash commandline does not display input and doesn't create newlines on enter.

akudan:
Fixes #49


master-of-zen:
Thanks for solution! 
That's problem with bash, on zsh systems everything works fine.

PR #51, opened 2020-04-24
Preserve ext

akudan:
Allows the user to specify the output file extension. Default behavior is the same if there is no extension or no output file specified.


master-of-zen:
Muxing should always be to .mkv container.
My testing showed showed that other containers like mp4, avi prone to errors.

akudan:
This doesn't affect muxing in any intermediate step. It should only affect the final output, which I think should be user's choice. MP4 spec for AV1 is more mature than MKV!

PR #56, opened 2020-05-06
Reduce Cognitive Complexity for code climate to increase maintainability

luigi311:
Refactoring functions to reduce cognitive complexity. ~This is still WIP so it is not ready, created PR so others can see progress/contribute.~ I am unable to completely reduce scene_detect to 5 or below so this is all set.

TODO:
- [X] check_executables
- [X] setup
- [x] process_input
- [ ] scene_detect
- [x] get_video_queue

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,9 @@
+version: "2"
+plugins:
+  fixme:
+    enabled: true
+checks:
+  file-lines:
+    enabled: false
+  method-lines:
+    enabled: false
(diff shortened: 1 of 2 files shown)


luigi311:
Here is what code climate is reporting with both check_executables and setup fixed, process_input is still WIP. 
[codeclimate.zip](https://github.com/master-of-zen/Av1an/files/4584900/codeclimate.zip)

luigi311:
Process_inputs is now simplified, i moved all the calls from arg_parsing to main thread following the arg_parsing call, i then did an if condition to check if server mode is being used to not even call process_input and still print out the ignore input if needed. This allows me to remove the nested if else conditions that were needed due to checking if its in server mode.

luigi311:
Ready to validation and merging

PR #64, opened 2020-05-15
Config file saved in JSON rather than pickle format for easy editing

farid-fari:
The goal of config files should be to allow users to create/edit them manually to avoid command-line clutter, so this will make this possible.


master-of-zen:
Ok, thanks :+1:

PR #65, opened 2020-05-15
Minor fixes

farid-fari:
Sometimes the container has the frame count and it's faster to just use that rather than count frames with `ffmpeg`. Requires `ffprobe` to be available, I'm not sure how common this is amongst `ffmpeg` installs.

See: https://stackoverflow.com/questions/2017843/fetch-frame-count-with-ffmpeg

Imports `subprocess.PIPE` and `subprocess.STDOUT` for cleaner code.

### DIFF
--- a/av1an.py
+++ b/av1an.py
@@ -83,32 +83,7 @@ class Av1an:
     @staticmethod
     def frame_probe(source: Path):
         """Get frame count."""
-
-        # First try using ffprobe (faster). See:
-        # https://stackoverflow.com/questions/2017843/fetch-frame-count-with-ffmpeg
-        cmd = ["ffprobe",
-               "-v", "error",
-               "-select_streams", "v:0",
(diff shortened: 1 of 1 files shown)


master-of-zen:
Originally i was using ffprobe for frame counting, later moved to ffmpeg, because it's way-way faster and accurate.
Here is release when move was done https://github.com/master-of-zen/Av1an/releases/tag/1.5
Problem with ffprobe that it's extra dependency(which usually needs to be installed separately), and not that fast.
Also current method always return real number of frames. Sometimes splitting or bad source report wrong frame count.
So please remove ffprobe part from frame_probe and i will merge :) :+1:  :1st_place_medal: 
Also join AV1 discord, if you will, we can communicate there https://discord.gg/2n5UMYK

farid-fari:
Duly noted, removed the whole `ffprobe` shenanigans and added some minor improvements.

PR #70, opened 2020-05-21
Smooth ETA and FPS calculation

### LINKED ISSUE #68: [Feature request] More "stable" fps & ETA reporting
As it is currently, the encoding speed fps and ETA can update very rapidly and vary wildly. At least for me, when using 12 workers, the ETA estimate isn't very useful because of this, as it fluctuates quickly between various values between 7 and 24 hours. It also updates at an inconsistent rate, somtimes multiple times a second, sometimes less than once a second.

I really have two suggestions:

* Have the estimate update at a constant rate, maybe once or twice a second
* Have the fps and ETA be some sort of an average accumulated over time, to make the values more stable. Maybe take into account the total average fps (total number of frames encoded divided by elapsed time) and use it with a more recent fps figure to calculate an average or a weighted average.

n9Mtq4:
Smooths the time remaining and fps calculation in the progress bar. Fixes #68.

Since the average is smoothed so much, we need to reset the progress bar on the first update, otherwise it takes about 1000 frames for the fps to reach its correct value after no updates during the first pass.

### DIFF
--- a/av1an.py
+++ b/av1an.py
@@ -51,9 +51,13 @@ def Manager():
 
 class Counter(object):
     def __init__(self, total, initial):
-        self.bar = tqdm(total=total, initial=initial, dynamic_ncols=True, unit="fr", leave=True)
+        self.first_update = True
+        self.bar = tqdm(total=total, initial=initial, dynamic_ncols=True, unit="fr", leave=True, smoothing=0.01)
 
     def update(self, value):
+        if self.first_update:
+            self.bar.reset()
+            self.first_update = False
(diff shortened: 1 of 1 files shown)


master-of-zen:
:+1:

master-of-zen:
@n9Mtq4 also it's breaking resume frame counter

master-of-zen:
https://github.com/master-of-zen/Av1an/commit/9514892b759dd5e644505f2dc711fd7ea2e8fe9c

PR #86, opened 2020-06-16
FIx for video split using aom_keyframes

Quantum1248:
Added -w and -h params to the command used to call the aomenc encoder when using aom_keyframes as splitting method. Not using those params cause an error in aomenc with some files.


master-of-zen:
Thank you :medal_military:

PR #87, opened 2020-06-16
Added support for ' in the input's filename

Quantum1248:
In the concat file, the files were listed as 

> file 'path'

so if in the path there was a ' the result were 

> file 'path1'path2'

and ffmpge would read only path1 and throw an error because it didn't find the file.
With this commits now all the ' in the path are replaced with '/'' so ffmpeg can read the path correctly.


master-of-zen:
@Quantum1248 :+1:

PR #88, opened 2020-06-17
Fix error where new was used before declaration

rHermes:
When running with the `rav1e` encoder, I would get an error saying that `new` was used before declaration.

Simply declaring it in the top is enough to not trigger the exception.


master-of-zen:
`if new > frame:` block was not intended properly.
Proper code is:
```
if match:
    new = int(match.group(1))
    if new > frame:
        counter.update(new - frame)
        frame = new
```
Thanks for pointing out :+1: 
Fixed in https://github.com/master-of-zen/Av1an/commit/c5b6632327f96a331981b2823e95c2ac3d84d41e
Refactoring a lot of stuff:smile:

PR #90, opened 2020-06-20
Add option for minimum number of frames in a chunk

n9Mtq4:
This PR works towards making splits less frequent than the encoder's key frame placement.

This PR adds the option `--min_scene_len` to specify the minimum number of frames in a chunk. This applies to scenedetect and aom_keyframe split methods. When this option is not set, the split behavior is unchanged.

Libaom's key frame placement was recently changed to include a lower bound on the distance between key frames (used by specifying`--kf-min-dist`) ([libaom commit](https://aomedia.googlesource.com/aom/+/ce97de2724d7ffdfdbe986a14d49366936187298)). This changes the way aom decides key frame placements. This PR updates the aom key frame split method to match the new aom code that accounts for this minimum distance.

### DIFF
--- a/av1an.py
+++ b/av1an.py
@@ -380,9 +380,10 @@ class Av1an:
         if split_method == 'pyscene':
             queue_fix = not self.d.get('queue')
             threshold = self.d.get("threshold")
-            self.log(f'Starting scene detection Threshold: {threshold}\n')
+            min_scene_len = self.d.get('min_scene_len')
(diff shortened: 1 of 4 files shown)


master-of-zen:
@n9Mtq4 tested to work as intended?

n9Mtq4:
Yep. Both split methods still cut on large changes and do not produce chunks smaller than the minimum length.

master-of-zen:
@n9Mtq4 :1st_place_medal: :)

PR #91, opened 2020-06-22
Use argparse for split method choice validation

n9Mtq4:
This PR uses the choices option in argparse for validating if the split method is correct. This allows for argparse to print an error so 3 lines in av1an.py checking for an invalid choice can be removed.

The new error message.
```
av1an.py: error: argument --split_method: invalid choice: '1324' (choose from 'pyscene', 'aom_keyframes')
```

It also prints out the choices in `--help`.
```
  --split_method {pyscene,aom_keyframes}
                        Specify splitting method
```

### DIFF
--- a/av1an.py
+++ b/av1an.py
@@ -342,9 +342,6 @@ class Av1an:
                 self.log('Error in aom_keyframes')
                 print('Error in aom_keyframes')
                 terminate()
-        else:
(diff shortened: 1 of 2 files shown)


master-of-zen:
:1st_place_medal: :)

PR #92, opened 2020-06-22
Code cleanup

n9Mtq4:
* Removes unneeded spaces
* Makes sure double line breaks between top level functions and classes
* Fixes indentation
* Ensures line breaks at the end of files


master-of-zen:
:+1:

PR #93, opened 2020-06-22
Fix indent and remove extra len

n9Mtq4:
I missed an indent of 8 and there was an extra len call on a log message.


master-of-zen:
:+1:

PR #94, opened 2020-06-22
Make aom keyframe split 1st pass match chunk 1st pass

n9Mtq4:
This PR makes the aom keyframe split method use similar params to the first pass run on chunks. This will allow for reusing the first pass stat file on the chunks in a later PR.

- [x] Process video params before the splitting
- [x] Generate aom first pass params for other encoders
- [x] Use video params and ffmpeg filters for aom keyframe split first pass

### DIFF
--- a/av1an.py
+++ b/av1an.py
@@ -49,6 +49,9 @@ class Av1an:
                 print(f'No such model: {Path(self.vmaf_path).as_posix()}')
                 terminate()
 
+        if self.video_params is None:
+            self.video_params = get_default_params_for_encoder(self.encoder)
+
     def target_vmaf(self, source):
         # TODO speed up for vmaf stuff
         # TODO reduce complexity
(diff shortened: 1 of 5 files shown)


n9Mtq4:
@Quantum1248 this PR overwrites your commit that adds `-w` and `-h` to `aom_keyframes`. With adding in the ffmpeg commands, any number of filters, (scale, crop, etc.) can change the video geometry. Instead of writing an ffmpeg filter parser, I expect the user to add `-w` and `-h` to `--video_params` if they need it for their video since they would already have added it for 2nd pass stuff too.

However, `--video_params` is only passed in to this split method if the encoder is aom since options for other encoders would cause a crash. I expect this wouldn't be a huge problem since one should use pyscenedetect with other encoders.

Does this work for your use case?

Quantum1248:
Yes I tried it and everything work.

master-of-zen:
@n9Mtq4 :+1: will check it out

PR #97, opened 2020-06-24
Reuse first pass from AOM keyframe split on the chunks

n9Mtq4:
This PR allows for the reuse of the first pass stats file generated in the aom_keyframes split method to be used on the chunks. This will save time from not redoing the first pass on each chunk and may allow for chunked VBR encoding.

- [x] Reuse argparse option
- [x] Change command composing
- [x] Hook up to encoding queue 
- [x] Segment first pass file
- [x] Recompute EOS packet
- [x] Make compatible with target vmaf and boosting
- [x] Determine if first frame stats needs to be recomputed - doesn't seem like it
- [x] Evaluate efficiency benefit or detriment - from my test clip the bitstream is identical to using a new first pass

### DIFF
--- a/av1an.py
+++ b/av1an.py
@@ -45,6 +45,10 @@ class Av1an:
                 print(f'No such model: {Path(self.vmaf_path).as_posix()}')
                 terminate()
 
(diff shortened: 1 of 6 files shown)


n9Mtq4:
@master-of-zen This is ready for your review.

PR #98, opened 2020-06-24
Fix using input instead of self.input and add as_posix on Paths

n9Mtq4:
A user on the discord reported a bug on Windows. There were some cases where Paths weren't converted to strings. It works fine on linux but causes problems on Windows. This also fixes passing the built in `input` function instead of the input path to a couple functions.


master-of-zen:
@n9Mtq4  :+1:

PR #99, opened 2020-06-24
Added lock to I/O operations on done.json

Quantum1248:
I added a lock for when a encoding process need to read/write the done.json file. Without it, it may happen that two process read the file before writing the result, and then the second one to write the result would overwrite the result of the first.


master-of-zen:
@Quantum1248 :+1:

PR #102, opened 2020-06-26
Simplify first pass reuse command generation

n9Mtq4:
This simplifies the logic so that first pass reuse modifies the commands directly instead of relying on aom composing function, boost, and target vmaf.


master-of-zen:
@n9Mtq4 :+1:

PR #104, opened 2020-07-02
Actions

luigi311:
This will add a github actions script that will test multiple different features to make sure a commit/pr is not breaking something.


master-of-zen:
@luigi311 :+1:

PR #108, opened 2020-07-25
VMAF calculation resolution

megapro17:


master-of-zen:
@megapro17 :medal_sports:

PR #111, opened 2020-08-01
Put vmaf log in destination folder & use Path instead of string

n9Mtq4:
This fixes [plonk420's issue](https://discord.com/channels/587033243702788123/686197633181810720/733181191770996756) reported on discord. 

`call_vmaf` puts the vmaf log file in the same directory as the source video. This works well for target vmaf, but it doesn't make sense for the final vmaf plot. It also causes problems if the user doesn't have write permissions to the directory the source video is in. This PR puts the log in the output directory as well as as small refactors to `call_vmaf` and `plot_vmaf`.

### DIFF
--- a/Av1an/vmaf.py
+++ b/Av1an/vmaf.py
@@ -28,7 +28,7 @@ def read_vmaf_json(file, percentile):
     return perc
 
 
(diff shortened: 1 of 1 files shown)


master-of-zen:
Good :medal_military:

PR #112, opened 2020-08-05
Use a chunk queue and access chunks with an ffmpeg command

n9Mtq4:
Currently Av1an uses chunk files as the basis for encoding. This PR moves to replace chunk files with chunk objects. These new chunks will be accessed by an ffmpeg command rather than the file itself. This simplifies the logic in many places and will allow for future PRs allowing for alternatives to segmenting chunks to disk.

Progress
- [x] Chunk class and queue
- [x] New command composing
- [x] Encode from ffmpeg command
- [x] Resuming
- [x] Target VMAF
- [x] First pass reuse
- [x] ~~Boost?~~
- [x] VVC
- [x] Testing


master-of-zen:
@n9Mtq4 :1st_place_medal:

PR #115, opened 2020-08-07
Add the option for chunks to use the select filter instead of segmenting to disk

n9Mtq4:
This PR introduces the option of using [ffmpeg's select filter](https://ffmpeg.org/ffmpeg-filters.html#select_002c-aselect) to create the chunks. This is enabled with `--chunk_method select`. This has the advantage of not requiring intermediate chunk files written to disk and being able to split on non-keyframes. The downside to this is increased time as ffmpeg needs to seek to the frames.

Performance on 53000 frame 1080p encode with libaom cpu-used 6:
segment: 4813.7s
select: 6853.3s

An extra 34 minutes, but this depends on the number of frames and becomes less important on lower cpu-used values. In addition, the select encode had an extra chunk because it was able to split on non-keyframes.

### DIFF
--- a/Av1an/arg_parse.py
+++ b/Av1an/arg_parse.py
@@ -12,6 +12,7 @@ class Args(object):
         self.output_file = None
 
         # Splitting
+        self.chunk_method = None
         self.scenes = None
         self.split_method = None
         self.extra_split = None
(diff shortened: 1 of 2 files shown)


Felixkruemel:
I personally would stay on the current method as the speed decrease here is roughly 30% which is insanely high.

The best thing would be an option for that. Some users have small drive space. For them that would be better.

n9Mtq4:
Well I have good news. The next PR will add a third option that can split on non-keyframes, doesn't require intermediate files, and is faster.

Felixkruemel:
@n9Mtq4 That's even better!

master-of-zen:
@n9Mtq4 Great job :1st_place_medal: :+1:

PR #116, opened 2020-08-09
Add the option for chunks to use vapoursynth + ffms2

n9Mtq4:
This PR introduces the option of using vapoursynth and ffms2 to create the chunks. This is enabled with `--chunk_method vs_ffms2`. This has the advantage of being quick, not requiring intermediate chunk files written to disk, and being able to split on non-keyframes. The only downside is the requirement of having vapoursynth and ffms2 installed.

Performance on 53000 frame 1080p encode with libaom cpu-used 6:
segment: 4813.7s
vs_ffms2: 4796.0s


master-of-zen:
@n9Mtq4 :+1:

PR #118, opened 2020-08-11
Fix target vmaf min/max q default setting

n9Mtq4:
This fixes a bug where if you wanted to set only the min q or only the max q, it would override your option with the default.


master-of-zen:
@n9Mtq4 :+1:

PR #119, opened 2020-08-12
Fix handling path with spaces

Fuchs4:
This fixes handling path with spaces in them.
av1an stores commands in strings, which are split before passing them to the subproces module. This causes any path with spaces in it to be split in multiple parts.
This patch changes all commands (i hope i found all) to be stored in tupel, which can be directly passed to the subprocess module, eliminating the need for splitting them.
For concat i have used shlex to quote paths, which can quote/escape most (if not all) unicode charactes correctly so that ffmpeg sees the correct path.

### DIFF
--- a/Av1an/bar.py
+++ b/Av1an/bar.py
@@ -62,7 +62,7 @@ def process_pipe(pipe):
 
 
 def make_vvc_pipe(command):
-    pipe = subprocess.Popen(command[0], stdout=PIPE,
+    pipe = subprocess.Popen(command, stdout=PIPE,
                             stderr=STDOUT,
                             universal_newlines=True)
     return pipe
--- a/Av1an/bar.py
+++ b/Av1an/bar.py
@@ -3,6 +3,7 @@
 import re
 import subprocess
 import sys
+from typing import Tuple
 from collections import deque
 from multiprocessing.managers import BaseManager
 from subprocess import PIPE, STDOUT
--- a/Av1an/bar.py
+++ b/Av1an/bar.py
@@ -34,13 +34,11 @@ class Counter():
 BaseManager.register('Counter', Counter)
 
 
-def make_pipes(ffmpeg_gen, command):
+def make_pipes(ffmpeg_gen_cmd: list, command: list):
 
-    f, e = command.split('|')
-    f, e = f.split(), e.split()
-    ffmpeg_gen_pipe = subprocess.Popen(ffmpeg_gen.split(), stdout=PIPE, stderr=STDOUT)
-    ffmpeg_pipe = subprocess.Popen(f, stdin=ffmpeg_gen_pipe.stdout, stdout=PIPE, stderr=STDOUT)
-    pipe = subprocess.Popen(e, stdin=ffmpeg_pipe.stdout, stdout=PIPE,
+    ffmpeg_gen_pipe = subprocess.Popen(ffmpeg_gen_cmd, stdout=PIPE, stderr=STDOUT)
+    ffmpeg_pipe = subprocess.Popen(command[0], stdin=ffmpeg_gen_pipe.stdout, stdout=PIPE, stderr=STDOUT)
+    pipe = subprocess.Popen(command[1], stdin=ffmpeg_pipe.stdout, stdout=PIPE,
                             stderr=STDOUT,
                             universal_newlines=True)
 
@@ -64,7 +62,7 @@ def process_pipe(pipe):
 
 
 def make_vvc_pipe(command):
-    pipe = subprocess.Popen(command.split(), stdout=PIPE,
+    pipe = subprocess.Popen(command[0], stdout=PIPE,
                             stderr=STDOUT,
                             universal_newlines=True)
     return pipe
--- a/Av1an/bar.py
+++ b/Av1an/bar.py
@@ -34,7 +34,7 @@ class Counter():
 BaseManager.register('Counter', Counter)
 
 
-def make_pipes(ffmpeg_gen_cmd: list, command: list):
+def make_pipes(ffmpeg_gen_cmd: Tuple[str], command: Tuple[Tuple[str]]):
 
     ffmpeg_gen_pipe = subprocess.Popen(ffmpeg_gen_cmd, stdout=PIPE, stderr=STDOUT)
     ffmpeg_pipe = subprocess.Popen(command[0], stdin=ffmpeg_gen_pipe.stdout, stdout=PIPE, stderr=STDOUT)
--- a/Av1an/chunk.py
+++ b/Av1an/chunk.py
@@ -11,7 +11,7 @@ class Chunk:
     to be run on this chunk.
     """
 
-    def __init__(self, temp: Path, index: int, ffmpeg_gen_cmd: str, output_ext: str, size: int, frames: int):
+    def __init__(self, temp: Path, index: int, ffmpeg_gen_cmd: list, output_ext: str, size: int, frames: int):
         """
         Chunk class constructor
 
--- a/Av1an/chunk.py
+++ b/Av1an/chunk.py
@@ -11,7 +11,7 @@ class Chunk:
     to be run on this chunk.
     """
 
-    def __init__(self, temp: Path, index: int, ffmpeg_gen_cmd: list, output_ext: str, size: int, frames: int):
+    def __init__(self, temp: Path, index: int, ffmpeg_gen_cmd: Tuple[str], output_ext: str, size: int, frames: int):
         """
         Chunk class constructor
 
--- a/Av1an/chunk.py
+++ b/Av1an/chunk.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Dict, Any
+from typing import Dict, Any, Tuple
 
 from .arg_parse import Args
 import Av1an
--- a/Av1an/chunk_queue.py
+++ b/Av1an/chunk_queue.py
@@ -127,7 +127,7 @@ def create_vsffms2_chunk(args: Args, index: int, load_script: Path, frame_start:
     frames = frame_end - frame_start
     frame_end -= 1  # the frame end boundary is actually a frame that should be included in the next chunk
 
-    ffmpeg_gen_cmd = f'vspipe {load_script} -y - -s {frame_start} -e {frame_end}'
+    ffmpeg_gen_cmd = ('vspipe', load_script, '-y', '-', '-s', frame_start, '-e', frame_end)
     extension = get_file_extension_for_encoder(args.encoder)
     size = frames  # use the number of frames to prioritize which chunks encode first, since we don't have file size
 
@@ -176,7 +176,7 @@ def create_select_chunk(args: Args, index: int, src_path: Path, frame_start: int
     frames = frame_end - frame_start
     frame_end -= 1  # the frame end boundary is actually a frame that should be included in the next chunk
 
-    ffmpeg_gen_cmd = f'ffmpeg -y -hide_banner -loglevel error -i {src_path.as_posix()} -vf select=between(n\\,{frame_start}\\,{frame_end}),setpts=PTS-STARTPTS {args.pix_format} -bufsize 50000K -f yuv4mpegpipe -'
+    ffmpeg_gen_cmd = ('ffmpeg', '-y', '-hide_banner', '-loglevel', 'error', '-i', src_path.as_posix(), '-vf', f'select=between(n\,{frame_start}\,{frame_end}),setpts=PTS-STARTPTS', *args.pix_format, '-bufsize', '50000K', '-f', 'yuv4mpegpipe', '-')
     extension = get_file_extension_for_encoder(args.encoder)
     size = frames  # use the number of frames to prioritize which chunks encode first, since we don't have file size
 
@@ -223,7 +223,7 @@ def create_chunk_from_segment(args: Args, index: int, file: Path) -> Chunk:
     :param file: the segmented file
     :return: A Chunk
     """
-    ffmpeg_gen_cmd = f'ffmpeg -y -hide_banner -loglevel error -i {file.as_posix()} {args.pix_format} -bufsize 50000K -f yuv4mpegpipe -'
+    ffmpeg_gen_cmd = ('ffmpeg', '-y', '-hide_banner', '-loglevel', 'error', '-i', file.as_posix(), *args.pix_format, '-bufsize', '50000K', '-f', 'yuv4mpegpipe', '-')
     file_size = file.stat().st_size
     frames = frame_probe(file)
     extension = get_file_extension_for_encoder(args.encoder)
--- a/Av1an/chunk_queue.py
+++ b/Av1an/chunk_queue.py
@@ -147,7 +147,10 @@ def create_video_queue_select(args: Args, split_locations: List[int]) -> List[Ch
     """
     # add first frame and last frame
     last_frame = frame_probe(args.input)
-    split_locs_fl = [0] + split_locations + [last_frame]
+    if 0 not in split_locations:
+        split_locs_fl = [0] + split_locations
+    if last_frame not in split_locations:
+        split_locations = split_locations + [last_frame]
 
     # pair up adjacent members of this list ex: [0, 10, 20, 30] -> [(0, 10), (10, 20), (20, 30)]
     chunk_boundaries = zip(split_locs_fl, split_locs_fl[1:])
--- a/Av1an/chunk_queue.py
+++ b/Av1an/chunk_queue.py
@@ -147,11 +147,6 @@ def create_video_queue_select(args: Args, split_locations: List[int]) -> List[Ch
     """
     # add first frame and last frame
     last_frame = frame_probe(args.input)
-    if 0 not in split_locations:
-        split_locs_fl = [0] + split_locations
-    if last_frame not in split_locations:
-        split_locations = split_locations + [last_frame]
-
     # pair up adjacent members of this list ex: [0, 10, 20, 30] -> [(0, 10), (10, 20), (20, 30)]
     chunk_boundaries = zip(split_locs_fl, split_locs_fl[1:])
 
--- a/Av1an/chunk_queue.py
+++ b/Av1an/chunk_queue.py
@@ -127,7 +127,7 @@ def create_vsffms2_chunk(args: Args, index: int, load_script: Path, frame_start:
     frames = frame_end - frame_start
     frame_end -= 1  # the frame end boundary is actually a frame that should be included in the next chunk
 
-    ffmpeg_gen_cmd = ('vspipe', load_script, '-y', '-', '-s', frame_start, '-e', frame_end)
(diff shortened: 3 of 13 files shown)


n9Mtq4:
Might it be possible to achieve the same functionality by using `shlex.split` and keeping the commands as strings internally?

Av1an/bar.py
```diff
@@ -34,13 +34,11 @@ def update(self, value):
 BaseManager.register('Counter', Counter)
 
 
-def make_pipes(ffmpeg_gen, command):
+def make_pipes(ffmpeg_gen_cmd: list, command: list):
```
n9Mtq4, line 37: `ffmpeg_gen_cmd` is now a `Tuple[str]`.

`command` is now a tuple containing a list and a tuple `(['ffmpeg', ...], ('aomenc', ...))`. `command` should contain only 1 type inside of it. Pick either `Tuple[List[str]]` or `Tuple[Tuple[str]]`. Since it seems like most commands are going to be tuples, I would prefer `Tuple[Tuple[str]]`.

```suggestion
def make_pipes(ffmpeg_gen_cmd: Tuple[str], command: Tuple[Tuple[str]]):
```

Av1an/bar.py
```diff
@@ -64,7 +62,7 @@ def process_pipe(pipe):
 
 
 def make_vvc_pipe(command):
-    pipe = subprocess.Popen(command.split(), stdout=PIPE,
+    pipe = subprocess.Popen(command[0], stdout=PIPE,
```
n9Mtq4, line 65: `command` here was only 1 string and is now only 1 tuple, so no indexing is needed.

```suggestion
    pipe = subprocess.Popen(command, stdout=PIPE,
```

Av1an/chunk_queue.py
```diff
@@ -127,7 +127,7 @@ def create_vsffms2_chunk(args: Args, index: int, load_script: Path, frame_start:
     frames = frame_end - frame_start
     frame_end -= 1  # the frame end boundary is actually a frame that should be included in the next chunk
 
-    ffmpeg_gen_cmd = f'vspipe {load_script} -y - -s {frame_start} -e {frame_end}'
+    ffmpeg_gen_cmd = ('vspipe', load_script, '-y', '-', '-s', frame_start, '-e', frame_end)
```
n9Mtq4, line 130: Some of these are not strings and now need to be explicitly converted to strings.

```suggestion
    ffmpeg_gen_cmd = ('vspipe', load_script.as_posix(), '-y', '-', '-s', str(frame_start), '-e', str(frame_end))
```

Av1an/chunk_queue.py
```diff
@@ -147,7 +147,10 @@ def create_video_queue_select(args: Args, split_locations: List[int]) -> List[Ch
     """
     # add first frame and last frame
     last_frame = frame_probe(args.input)
-    split_locs_fl = [0] + split_locations + [last_frame]
+    if 0 not in split_locations:
+        split_locs_fl = [0] + split_locations
+    if last_frame not in split_locations:
+        split_locations = split_locations + [last_frame]
```
n9Mtq4, line 153: Why is this needed? `split_location` will never have the first or last frame included.

Av1an/compose.py
```diff
@@ -9,7 +9,7 @@
 from .vvc import get_yuv_file_path
 
 
-def gen_pass_commands(args: Args, chunk: Chunk) -> List[str]:
+def gen_pass_commands(args: Args, chunk: Chunk) -> List[List]:
```
n9Mtq4, line 12: This return type is wrong.
```suggestion
def gen_pass_commands(args: Args, chunk: Chunk) -> Tuple[Tuple[str]]:
```

Av1an/chunk.py
```diff
@@ -11,7 +11,7 @@ class Chunk:
     to be run on this chunk.
     """
 
-    def __init__(self, temp: Path, index: int, ffmpeg_gen_cmd: str, output_ext: str, size: int, frames: int):
+    def __init__(self, temp: Path, index: int, ffmpeg_gen_cmd: list, output_ext: str, size: int, frames: int):
```
n9Mtq4, line 14:
```suggestion
    def __init__(self, temp: Path, index: int, ffmpeg_gen_cmd: Tuple[str], output_ext: str, size: int, frames: int):
```

Av1an/compose.py
```diff
@@ -86,9 +86,9 @@ def compose_ffmpeg_pipe(a: Args) -> str:
     Gets the ffmpeg command with filters that pipes into the encoder
 
     :param a: the Args
-    :return: the ffmpeg command as a string
+    :return: the ffmpeg command as a list
     """
-    return f'ffmpeg -y -hide_banner -loglevel error -i - {a.ffmpeg} {a.pix_format} -bufsize 50000K -f yuv4mpegpipe -'
+    return ['ffmpeg', '-y', '-hide_banner', '-loglevel', 'error', '-i', '-', *a.ffmpeg_pipe]
```
n9Mtq4, line 91: Since every other command is a tuple, this one should be too. You should also update the return type in the function signature.
```suggestion
    return ('ffmpeg', '-y', '-hide_banner', '-loglevel', 'error', '-i', '-', *a.ffmpeg_pipe)
```

Av1an/compose.py
```diff
@@ -86,9 +86,9 @@ def compose_ffmpeg_pipe(a: Args) -> str:
...
 
 
 def compose_svt_av1(a: Args, c: Chunk) -> List[str]:
```
n9Mtq4, line 94: Update signature.
```suggestion
def compose_svt_av1(a: Args, c: Chunk) -> Tuple[Tuple[str]]:
```

Av1an/compose.py
```diff
@@ -100,14 +100,14 @@ def compose_svt_av1(a: Args, c: Chunk) -> List[str]:
...
-            f'{compose_ffmpeg_pipe(a)} | SvtAv1EncApp -i stdin {a.video_params} -output-stat-file {c.fpf}.stat -b {os.devnull} -',
-            f'{compose_ffmpeg_pipe(a)} | SvtAv1EncApp -i stdin {a.video_params} -input-stat-file {c.fpf}.stat -b {c.output} -',
-        ]
+        return (
+            (compose_ffmpeg_pipe(a), ('SvtAv1EncApp', '-i', 'stdin', *a.video_params, '-output-stat-file', f'{c.fpf}.stat', '-b', os.devnull, '-')),
+            (compose_ffmpeg_pipe(a), ('SvtAv1EncApp', '-i', 'stdin', *a.video_params, '-input-stat-file', f'{c.fpf}.stat', '-b', c.output, '-'))
+        )
 
 
 def compose_rav1e(a: Args, c: Chunk) -> List[str]:
```
n9Mtq4, line 113:
```suggestion
def compose_rav1e(a: Args, c: Chunk) -> Tuple[Tuple[str]]:
```

Av1an/compose.py
```diff
@@ -120,14 +120,14 @@ def compose_rav1e(a: Args, c: Chunk) -> List[str]:
...
-            f'{compose_ffmpeg_pipe(a)} | rav1e - --first-pass {c.fpf}.stat {a.video_params} --output {os.devnull}',
-            f'{compose_ffmpeg_pipe(a)} | rav1e - --second-pass {c.fpf}.stat {a.video_params} --output {c.output}',
-        ]
+        return (
+            (compose_ffmpeg_pipe(a), ('rav1e', '-', '--first-pass', f'{c.fpf}.stat', *a.video_params, '--output', os.devnull)),
+            (compose_ffmpeg_pipe(a), ('rav1e', '-', '--second-pass', f'{c.fpf}.stat', *a.video_params, '--output', c.output))
+        )
 
 
 def compose_aom(a: Args, c: Chunk) -> List[str]:
```
n9Mtq4, line 133:
```suggestion
def compose_aom(a: Args, c: Chunk) -> Tuple[Tuple[str]]:
```

Av1an/compose.py
```diff
@@ -139,14 +139,14 @@ def compose_aom(a: Args, c: Chunk) -> List[str]:
...
-            f'{compose_ffmpeg_pipe(a)} | aomenc --passes=2 --pass=1 {a.video_params} --fpf={c.fpf}.log -o {os.devnull} -',
-            f'{compose_ffmpeg_pipe(a)} | aomenc --passes=2 --pass=2 {a.video_params} --fpf={c.fpf}.log -o {c.output} -',
-        ]
+        return (
+            (compose_ffmpeg_pipe(a), ('aomenc', '--passes=2', '--pass=1', *a.video_params, f'--fpf={c.fpf}.log', '-o', os.devnull, '-')),
+            (compose_ffmpeg_pipe(a), ('aomenc', '--passes=2', '--pass=2', *a.video_params, f'--fpf={c.fpf}.log', '-o', c.output, '-'))
+        )
 
 
 def compose_vpx(a: Args, c: Chunk) -> List[str]:
```
n9Mtq4, line 152:
```suggestion
def compose_vpx(a: Args, c: Chunk) -> Tuple[Tuple[str]]:
```

Av1an/compose.py
```diff
@@ -158,14 +158,14 @@ def compose_vpx(a: Args, c: Chunk) -> List[str]:
...
-            f'{compose_ffmpeg_pipe(a)} | vpxenc --passes=2 --pass=1 {a.video_params} --fpf={c.fpf} -o {os.devnull} -',
-            f'{compose_ffmpeg_pipe(a)} | vpxenc --passes=2 --pass=2 {a.video_params} --fpf={c.fpf} -o {c.output} -',
-        ]
+        return (
+            (compose_ffmpeg_pipe(a), ('vpxenc', '--passes=2', '--pass=1', *a.video_params, f'--fpf={c.fpf}', '-o', os.devnull, '-')),
+            (compose_ffmpeg_pipe(a), ('vpxenc', '--passes=2', '--pass=2', *a.video_params, f'--fpf={c.fpf}', '-o', c.output, '-'))
+        )
 
 
 def compose_x265(a: Args, c: Chunk) -> List[str]:
```
n9Mtq4, line 171:
```suggestion
def compose_x265(a: Args, c: Chunk) -> Tuple[Tuple[str]]:
```

Av1an/compose.py
```diff
@@ -177,14 +177,14 @@ def compose_x265(a: Args, c: Chunk) -> List[str]:
...
-            f'{compose_ffmpeg_pipe(a)} | x265 --log-level error --pass 1 --y4m {a.video_params} --stats {c.fpf}.log - -o {os.devnull}',
-            f'{compose_ffmpeg_pipe(a)} | x265 --log-level error --pass 2 --y4m {a.video_params} --stats {c.fpf}.log - -o {c.output}',
-        ]
+        return (
+            (compose_ffmpeg_pipe(a), ('x265', '--log-level', 'error', '--pass', '1', '--y4m', *a.video_params, '--stats', f'{c.fpf}.log', '-', '-o', os.devnull)),
+            (compose_ffmpeg_pipe(a), ('x265', '--log-level', 'error', '--pass', '2', '--y4m', *a.video_params, '--stats', f'{c.fpf}.log', '-', '-o', c.output))
+        )
 
 
 def compose_x264(a: Args, c: Chunk) -> List[str]:
```
n9Mtq4, line 190:
```suggestion
def compose_x264(a: Args, c: Chunk) -> Tuple[Tuple[str]]:
```

Av1an/compose.py
```diff
@@ -196,14 +196,14 @@ def compose_x264(a: Args, c: Chunk) -> List[str]:
...
-            f'{compose_ffmpeg_pipe(a)} | x264 --stitchable --log-level error --pass 1 --demuxer y4m {a.video_params} - --stats {c.fpf}.log - -o {os.devnull}',
-            f'{compose_ffmpeg_pipe(a)} | x264 --stitchable --log-level error --pass 2 --demuxer y4m {a.video_params} - --stats {c.fpf}.log - -o {c.output}',
-        ]
+        return (
+            (compose_ffmpeg_pipe(a), ('x264', '--stitchable', '--log-level', 'error', '--pass', '1', '--demuxer', 'y4m', *a.video_params, '-', '--stats', f'{c.fpf}.log', '-', '-o', os.devnull)),
+            (compose_ffmpeg_pipe(a), ('x264', '--stitchable', '--log-level', 'error', '--pass', '2', '--demuxer', 'y4m', *a.video_params, '-', '--stats', f'{c.fpf}.log', '-', '-o', c.output))
+        )
 
 
 def compose_vvc(a: Args, c: Chunk) -> List[str]:
```
n9Mtq4, line 209:
```suggestion
def compose_vvc(a: Args, c: Chunk) -> Tuple[Tuple[str]]:
```

Av1an/compose.py
```diff
@@ -215,9 +215,9 @@ def compose_vvc(a: Args, c: Chunk) -> List[str]:
     :return: A list of commands
     """
     yuv_file = get_yuv_file_path(c).as_posix()
-    return [
-        f'vvc_encoder -c {a.vvc_conf} -i {yuv_file} {a.video_params} -f {c.frames} --InputBitDepth=10 --OutputBitDepth=10 -b {c.output}',
-    ]
+    return (
+        ('vvc_encoder', '-c', a.vvc_conf, '-i', yuv_file, *a.video_params, '-f', c.frames, '--InputBitDepth=10', '--OutputBitDepth=10', '-b', c.output),
```
n9Mtq4, line 219: Need an explicit string conversion.
```suggestion
        ('vvc_encoder', '-c', a.vvc_conf, '-i', yuv_file, *a.video_params, '-f', str(c.frames), '--InputBitDepth=10', '--OutputBitDepth=10', '-b', c.output),
```

Av1an/utils.py
```diff
@@ -45,25 +45,31 @@ def get_cq(command):
     return int(matches[-1])
 
 
-def man_q(command: str, q: int):
+def man_q(command: list, q: int):
```
n9Mtq4, line 48:
```suggestion
def man_q(command: Tuple[str], q: int):
```

Av1an/concat.py
```diff
@@ -60,26 +61,24 @@ def concatenate_video(temp: Path, output, encoder: str):
...
-        audio = ''
+        audio = ()
 
     if encoder == 'x265':
 
-        cmd = f' ffmpeg -y -fflags +genpts  -hide_banner -loglevel error -f concat -safe 0 -i {temp / "concat"} ' \
-            f'{audio} -c copy -movflags frag_keyframe+empty_moov -map 0  -f mp4 - | ffmpeg -y -hide_banner -loglevel error -i - -c copy {output} '
-        concat = subprocess.run(cmd, shell=True, stdout=PIPE, stderr=STDOUT).stdout
+        cmd = ('ffmpeg', '-y', '-fflags', '+genpts', '-hide_banner', '-loglevel', 'error', '-f', 'concat', '-safe', '0', '-i', temp / "concat", *audio, '-c', 'copy', '-movflags', 'frag_keyframe+empty_moov', '-map', '0', '-f', 'mp4', output)
+        concat = subprocess.run(cmd, shell=False, stdout=PIPE, stderr=STDOUT).stdout
```
n9Mtq4, line 76: `shell=False` is the default, so just remove it.
```suggestion
        concat = subprocess.run(cmd, stdout=PIPE, stderr=STDOUT).stdout
```

Av1an/concat.py
```diff
@@ -60,26 +61,24 @@ def concatenate_video(temp: Path, output, encoder: str):
...
 
     else:
-        cmd = f' ffmpeg -y -hide_banner -loglevel error -f concat -safe 0 -i {temp / "concat"} ' \
-            f'{audio} -c copy -map 0  -y "{output}"'
+        cmd = ('ffmpeg', '-y', '-hide_banner', '-loglevel', 'error', '-f', 'concat', '-safe', '0', '-i', temp / "concat", *audio, '-c', 'copy', '-map', '0', output)
 
-        concat = subprocess.run(cmd, shell=True, stdout=PIPE, stderr=STDOUT).stdout
+        concat = subprocess.run(cmd, shell=False, stdout=PIPE, stderr=STDOUT).stdout
```
n9Mtq4, line 81:
```suggestion
        concat = subprocess.run(cmd, stdout=PIPE, stderr=STDOUT).stdout
```

Av1an/vmaf.py
```diff
@@ -54,20 +56,22 @@ def call_vmaf(chunk: Chunk, encoded: Path, n_threads, model, res, fl_path: Path
...
-    vmaf_filter = f"[distorted][ref]libvmaf=log_fmt='json':eof_action=endall:log_path={fl}{mod}{n_threads}\" -f null - "
+    cmd_out = ('-f', 'null', '-')
 
-    cmd = cmd + filter_complex + distorted + ref + vmaf_filter
+    cmd = (*cmd_in, *filter_complex, distorted + ref + vmaf_filter, *cmd_out)
 
-    ffmpeg_gen_pipe = subprocess.Popen(chunk.ffmpeg_gen_cmd.split(), stdout=PIPE, stderr=STDOUT)
-    pipe = subprocess.Popen(cmd, stdin=ffmpeg_gen_pipe.stdout, stdout=PIPE, stderr=STDOUT, shell=True, universal_newlines=True)
+    ffmpeg_gen_pipe = subprocess.Popen(chunk.ffmpeg_gen_cmd, stdout=PIPE, stderr=STDOUT)
+    pipe = subprocess.Popen(cmd, stdin=ffmpeg_gen_pipe.stdout, stdout=PIPE, stderr=STDOUT, shell=False, universal_newlines=True)
```
n9Mtq4, line 74:
```suggestion
    pipe = subprocess.Popen(cmd, stdin=ffmpeg_gen_pipe.stdout, stdout=PIPE, stderr=STDOUT, universal_newlines=True)
```

Av1an/chunk_queue.py
Fuchs4: I am not sure about the first frame, but the last frame gets added by extra_split when using --extra_split option, at least for some of my videos. I added it as a quick fix and it is my first time working with github and pull requests, it was not supposed to be included in this pull request, and I noticed it too late. Sorry for that. We should probably look at extra_split for this issue.

master-of-zen: For sure extra splits should not exists in a way it is currently

Av1an/bar.py
Fuchs4: compose_vvc creates a Tuple[Tuple[str]], like the other compose functions. I did this to be consistent with the other compose functions and to be prepared if we want to pipe from ffmpeg later. I do not know about vvc, but I guess it does not support reading from pipe (yet?). If you prefer using a Tuple[str] you also need to change compose_vvc.

Fuchs4: Sorry, you were right. I intended it to be a `Tuple[Tuple[str]]`, but looks like I missed a pair of brackets. It is all right with your change.

master-of-zen:
@Fuchs4 Thanks! :+1: 
@n9Mtq4 Great work!  :hugs:

Av1an/chunk_queue.py
n9Mtq4: Ok. We should take a look at extra splits. The reason I asked about this is that this is not the only place that the first and last frame is added to the split locations, so the other place might need changing as well.

PR #121, opened 2020-08-13
Refactor command types and encoder composing

n9Mtq4:
This refactors the commands and command composing structure.

 - Commands have type aliases (Use `MPCommands` instead of `Tuple[Tuple[Tuple[str, ...], ...], ...] <- the actual type hint before`)
 - Commands were changed from tuples to lists ([reasoning](https://stackoverflow.com/a/626871/5196460)).
 - Encoders have their own classes for generating commands and executable location. We can move more logic (q replacement, vvc yuv creation) to these classes in the future.


master-of-zen:
@n9Mtq4 :hugs:

PR #129, opened 2020-08-19
Fix failing x265 test

n9Mtq4:
Recently, Bitbucket stopped supporting Mercurial repositories. x265 didn't migrate in time, so their repository can no longer be accessed until they migrate to git. This PR switches from mercurial on bitbucket to git on github for installing x265.


master-of-zen:
:+1:

PR #130, opened 2020-08-19
Move more encoder specific logic to encoder classes.

Fuchs4:
Ideally everything encoder specific should be included in the encoder classes, so that the rest of the program can treat every encoder in the same way and does not need to care which encoder is used.

todo:
- move match_aom/vpx/rav1e ... from bar to encoder
- probably some fixes and cleanups I missed.

### DIFF
--- a/Av1an/encoders/aom.py
+++ b/Av1an/encoders/aom.py
@@ -2,8 +2,9 @@ import os
 
 from Av1an.arg_parse import Args
 from Av1an.chunk import Chunk
-from Av1an.commandtypes import MPCommands, CommandPair
+from Av1an.commandtypes import MPCommands, CommandPair, Command
 from Av1an.encoders.encoder import Encoder
+from Av1an.utils import list_index_of_regex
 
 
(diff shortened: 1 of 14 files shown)


n9Mtq4:
FYI, I have been working on the same thing. I have already moved all default options and vvc stuff into the encoder class, so you don't need to worry about that. https://github.com/master-of-zen/Av1an/compare/master...n9Mtq4:refactor/encoder

Av1an/encoders/aom.py
```diff
@@ -15,22 +16,36 @@ def __init__(self):
             output_extension='ivf'
         )
 
-    def compose_1_pass(self, a: Args, c: Chunk) -> MPCommands:
+    def compose_1_pass(self, a: Args, c: Chunk, output=None) -> MPCommands:
+        if not output:
+            output = c.output
```
n9Mtq4, line 21: Instead of putting this in every compose method, we should make output required and just pass c.output when needed.

PR #135, opened 2020-08-24
Map all audio tracks in extract_audio()

Fuchs4:
Map all audio and subtitle tracks for the audio file instead of just the first (ffmpeg default).


master-of-zen:
@Fuchs4 :+1:

PR #137, opened 2020-08-25
Support SVT-AV1 -q for target vmaf

n9Mtq4:
SVT-AV1 allows for setting the q value with `--qp` or `-q`, but target vmaf was only looking for `--qp`.


master-of-zen:
@n9Mtq4 👍

PR #138, opened 2020-08-26
Move default options and vvc yuv creation into encoder classes

n9Mtq4:
Moves default encoder options (default passes and default target vmaf range) into the encoder classes.
Also moves specific VVC behavior (yuv file creation and deletion) into more general before and after chunk encoding events.


master-of-zen:
@n9Mtq4 :medal_sports:

PR #142, opened 2020-09-04
use log to improve linearity

krakow10:
This transformation of vmaf values maps more linearly to cq values, which should improve the search accuracy.

vmaf vs cq

transformed vmaf vs cq


krakow10:
I didn't make a branch and I want to do 2 PRs

PR #143, opened 2020-09-04
Vmaf transform

krakow10:
This transformation of vmaf values maps more linearly to cq values, which should improve the search accuracy.

vmaf vs cq:
vmaf vc cq
transformed vmaf vs cq:
transformed vmaf vs cq

### DIFF
--- a/Av1an/VMAF/target_vmaf.py
+++ b/Av1an/VMAF/target_vmaf.py
@@ -210,8 +210,8 @@ def weighted_search(num1, vmaf1, num2, vmaf2, target):
     :return: Q for new probe
     """
 
-    dif1 = abs(-math.log(1-target/100) - (-math.log(1-vmaf2/100))
-    dif2 = abs(-math.log(1-target/100) - (-math.log(1-vmaf1/100))
+    dif1 = abs(-math.log(1-target/100) - (-math.log(1-vmaf2/100)))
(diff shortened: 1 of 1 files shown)


Av1an/VMAF/target_vmaf.py
```diff
@@ -210,8 +211,8 @@ def weighted_search(num1, vmaf1, num2, vmaf2, target):
     :return: Q for new probe
     """
 
-    dif1 = abs(target - vmaf2)
-    dif2 = abs(target - vmaf1)
+    dif1 = abs(-ln(1-target/100) - (-ln(1-vmaf2/100)))
+    dif2 = abs(-ln(1-target/100) - (-ln(1-vmaf1/100)))
```
n9Mtq4, line 215: @krakow10 
`vmaf1` and `vmaf2` could be 100, in which case you can't take the log. Can you add something to protect against this? Something like this should work
```
# ensure vmaf != 100, to prevent ln(0)
vmaf1 = min(vmaf1, 99.99)
vmaf2 = min(vmaf2, 99.99)
```

n9Mtq4:
Old:
Total Probes: 1636
RMSE: 13.714383197625907
Averge diff: 5.364736332818936
Number of cqs wrong: 40

This PR:
Total Probes: 1639
RMSE: 13.714222931631454
Averge diff: 5.362816854190615
Number of cqs wrong: 38

PR #144, opened 2020-09-04
Fix search algorithm

krakow10:
Redesigned the search to use upper and lower bounds variables.  Bad case happens for example when the target vmaf is between vmaf_cq[-3] and vmaf_cq[-1] but it's searching between vmaf_cq[-2] and vmaf_cq[-1].  Also cleaned the next_q dead code in the main loop.  I've never used python so I don't know how to test this code


n9Mtq4:
Thanks for these. I'll work on testing it later today.

master-of-zen:
@krakow10  :+1:

PR #145, opened 2020-09-04
Provide guesswork math for 1-2 probe vmaf target

krakow10:
Needs ln from #143 and transform_vmaf from #146, so merge those first.


master-of-zen:
@krakow10 :+1:

PR #154, opened 2020-09-22
Add live updating progress bar for svt-av1

n9Mtq4:
Uses SVT-AV1's new `--progress 2` mode to support updating the progress bar.


master-of-zen:
@n9Mtq4 :medal_sports:

PR #164, opened 2020-10-04
Actions: Check if encoded file exists + misc

luigi311:
Added a python requirement to the actions to prevent a warning message.
Moved the matrix flag to the end so things like the baseline wont randomly add spaces in between flags and you also know exactly where to look for the flags its testing
Added du on the output of av1an to validate that the encode did happen which will catch issues like x265 encoding failing but not reporting an error on the CI because the exit code didnt change from 0
Replace the deprecated set-env function with the new github env

x265 is now built into the base image so we don't need to build it here anymore

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ on:
 
 env:
   deps: python3-dev libglib2.0-0 libsm6 libxext6 libxrender-dev libgl1-mesa-glx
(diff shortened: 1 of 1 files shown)


luigi311:
Rebased to fix x265 runs with the new fixes in master

luigi311:
Need one more rebase once #170 is merged. Should fix all the issues with this one

luigi311:
@master-of-zen This is now good to go

master-of-zen:
@luigi311 :+1:

PR #167, opened 2020-10-04
Fix x265 commands

luigi311:
x265 was failing because the video parameter list was only discarding digits but not string arguments. So it was doing a check on ["-p","slow","--crf"] so when it got to slow and checked if it was a validate argument for x265 it was failing. This now changes it to generate the list with everything that starts with - which should be all the flags and leave out any strings or numbers. 

Changes the exit code to 1 if not using --force to report that there was an error.


master-of-zen:
@luigi311 :medal_sports:

PR #168, opened 2020-10-04
Fix output double suffix, default will append _{encoder} instead of _av1

luigi311:
Fix double suffix issue and if no output is provided it will now use _{encoder} instead of always appending _av1 even when not encoding in av1.


master-of-zen:
Cool)

master-of-zen:
@luigi311 :medal_military:

PR #170, opened 2020-10-05
Support output files in different folders

n9Mtq4:
`args.output_file.stem` incorrectly removed any paths before the file name (`folder/file` would become just `file`).


luigi311:
This is a better solution as it actually works correctly.

master-of-zen:
Cool

master-of-zen:
@n9Mtq4 :medal_military:

PR #174, opened 2020-10-10
Not use shell for mkvmerge concat

wwww-wwww:


master-of-zen:
@wwww-wwww :1st_place_medal:

PR #181, opened 2020-10-18
Pass frames to x265 for a given chunk.

adworacz:
Per the x265 documentation, if `--frames` is specified, the rate control
can use the information for better encoding decisions.

In addition, we should be encoding an exact number of frames anyways.

This might benefit x264 as well, but I'm not sure if that's guarranteed.
It's trivial to add, so it may be worth adding anyways.

See docs around `--frames` option: https://x265.readthedocs.io/en/master/cli.html#input-output-file-options

Testing note: I tested what happens with something like vspipe returns
less than the specified number of frames and things work "just fine" -
the encoder exits with no error message and the stream isn't corrupted.

### DIFF
--- a/Encoders/x265.py
+++ b/Encoders/x265.py
@@ -24,7 +24,7 @@ class X265(Encoder):
         return [
             CommandPair(
                 Encoder.compose_ffmpeg_pipe(a),
-                ['x265', '--y4m', *a.video_params, '-', '-o', output]
+                ['x265', '--y4m', '--frames', str(c.frames), *a.video_params, '-', '-o', output]
             )
         ]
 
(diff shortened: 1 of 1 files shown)


Encoders/x265.py
```diff
@@ -62,4 +62,4 @@ def match_line(self, line: str):
         :param line: one line of text output from the encoder
         :return: match object from re.search matching the number of encoded frames"""
 
-        return re.search(r"^(\d+)", line)
+        return re.search(r"^\[.*\]\s(\d+)\/\d+", line)
```
adworacz, line 65: For reference, the old x265 line output looked like:
```
221 frames
```

And the new output (when frames are specified) looks like:
```
[6.0%] 60/1000 frames, 29.98 fps, 945.95 kb/s, eta 0:00:31
```

master-of-zen:
@adworacz :medal_military:

PR #182, opened 2020-10-19
Fix SVT-AV1 two-pass command syntax

as-com:
This should fix a command syntax error I encountered when encoding two-pass with SVT-AV1.

### DIFF
--- a/Encoders/svtav1.py
+++ b/Encoders/svtav1.py
@@ -33,12 +33,12 @@ class SvtAv1(Encoder):
             CommandPair(
                 Encoder.compose_ffmpeg_pipe(a),
                 ['SvtAv1EncApp', '-i', 'stdin', '--progress', '2', '--irefresh-type', '2', *a.video_params,
-                 '--output-stat-file', f'{c.fpf}.stat', '-b', os.devnull, '-']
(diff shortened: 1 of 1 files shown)


Felixkruemel:
While a fix would be good, is there any reason why you prefer 2pass on SVT? It's not as efficient as 1pass currently. @as-com

as-com:
I'm trying to encode video with a target bitrate, and two-pass is always more efficient (unless this assumption doesn't hold true for SVT-AV1?).

Felixkruemel:
@as-com 2pass is not more efficient for SVT-AV1. At least it wasn't two months ago. The test actually stated that it was less efficient, that's the reason why the default value is 1pass on SVT.

SVT also supports crf on 1pass when you enable --enable-tpl-la 1 and set -q accordingly. You shouldn't use a target bitrate if you don't need a specific file size. Setting a quality parameter and let the encoder adjust the bitrate accordingly (light scenes need less bitrate than dark scenes) is way more efficient.

as-com:
Alright, this patch is verified to work on my machine; should be safe to merge.

master-of-zen:
@as-com Thank you for your time and effort! :medal_military:

Issue #187, opened 2020-11-03
WindowsPath is not iterable


TimeForANinja:
Version: Av1an 1.13.post9
Source: pip
System: Windows 10
Command: `av1an.exe --resume --split_method aom_keyframes -i in.mkv -enc aom -p 2 -v " --cpu-used=3 --bit-depth=8 --end-usage=vbr --target-bitrate=2000 " -w 8 -ff " -vf scale=-8:720 " -log av1an.log.txt -o output.mkv`
Console-Output:
```
Queue: 0 Workers: 0 Passes: 2
Params: --cpu-used=3 --bit-depth=8 --end-usage=vbr --target-bitrate=2000
0fr [00:00, ?fr/s]
Concatenation failed, error
At line: 27
Error:argument of type 'WindowsPath' is not iterable
```
Logfile:
```
Av1an Started
Command:
C:\Users\space\AppData\Local\Programs\Python\Python37\Scripts\av1an --resume --split_method aom_keyframes -i in.mkv -enc aom -p 2 -v  --cpu-used=3 --bit-depth=8 --end-usage=vbr --target-bitrate=2000  -w 8 -ff  -vf scale=-8:720  -log av1an.log.txt -o output.mkv
09:18:15 Resuming...
09:18:15 Resumed with 1686 encoded clips done

09:18:16 Concatenating
09:18:16 Concatenation failed, aborting, error: argument of type 'WindowsPath' is not iterable
09:20:43 
```

might be related to that 131-issue (don't wanna tag it yet)

master-of-zen:
@TimeForANinja what  happens if you remove temp folder and run same command again?

TimeForANinja:
takes a bit but same error

master-of-zen:
What only `av1an.exe -i in.mkv` do?
Also try same on file in test.zip
[test.zip](https://github.com/master-of-zen/Av1an/files/5480690/test.zip)

TimeForANinja:
tryed that `av1an.exe -i dota_test.mkv` and got the same `argument of type 'WindowsPath' is not iterable`

can try with my original clip but its a 2h movie so that gonna take a bit

master-of-zen:
Have you tried to call av1an in the same folder as the file?

TimeForANinja:
> Have you tried to call av1an in the same folder as the file?

did just do it
copied the bin from the pip install folder and tryed `.\av1an.exe -i .\dota_test.mkv`
same error, same line

[...]

PR #188, opened 2020-11-03
Add as_posix to ffmpeg concat command

n9Mtq4:
I don't have access to Windows to test if I found all cases of missing `as_posix`, but this should fixe #187.


master-of-zen:
@n9Mtq4 :+1:

PR #189, opened 2020-11-08
Libav1an/av1an split allows guis to use av1an

natis1:
This commit does two things: First, it puts the core/backend functionality of av1an into its own package. Second, it removes ALL graphical dependencies from this backend, placing the responsibility of having them ON  the frontend.

The av1an executable is now in ./Av1an/av1an.py


master-of-zen:
Ugly and useless

PR #190, opened 2020-11-10
fix cli arguments in readme

Snoop05:


master-of-zen:
@Snoop05 :1st_place_medal:

PR #194, opened 2020-11-26
remove fuzzywuzzy/levenshtein and use difflib sequencematcher

wwww-wwww:
I also found it to be more accurate for this use case

levenshtein:
'--kxf-max-dusst' isn't a valid param for aom. Did you mean '--qm-max'?

sequencematcher:
'--kxf-max-dusst' isn't a valid param for aom. Did you mean '--kf-max-dist'?


master-of-zen:
@wwww-wwww :medal_sports:  A-a-amazing

PR #195, opened 2020-11-26
Add testing for vmaf and vmaf_plots, move non target_quality specific functions to vmaf.py

luigi311:
This will add testing for the vmaf and vmaf_plots arguments since they are currently broken and gets changed every so often.

I also moved generic vmaf functions out of the tartget_quality file since they do not belong there.

Action will also use tree -a instead of ls -a so a more complete output is shown

vmaf and vmaf_plots are currently broken due to the get_percentile function not working


master-of-zen:
@luigi311 :medal_sports:

PR #197, opened 2020-12-13
input: to be able to have several directories and files

sybilin:
Currently in the inputs list if a directory is in first position the following inputs are ignored. The PR allows to have several directories or files in input, in any order. All input files and files in directories will be taken into account.

For example if we have : -i dir1 file1 dir2 file2
the files taken into account will be file1, file2 and all relevant files from dir1 and dir2


master-of-zen:
@sybilin Thanks :1st_place_medal:

PR #198, opened 2020-12-14
remove unnecessary proj.outputs_filenames()

sybilin:
The call to proj.outputs_filenames() is useless since it is already before called in create_project_list()


master-of-zen:
@sybilin :1st_place_medal:

PR #199, opened 2020-12-15
remove unnecessary write_scenes_to_file

sybilin:
Those write_scenes_to_file are never called since the case with project.scenes is handled in the previous if in split_routine


master-of-zen:
@sybilin :medal_sports:

PR #205, opened 2020-12-24
Actions: Test baseline and target_quality for all encoders

luigi311:
Update the matrix to test baseline and target_quality on all encoders but svt-vp9 and vvc. To add more flags that needs to be tested on all encoders you need to add it name:[] and then create an include for it with just the same name and the flag for it.


master-of-zen:
@luigi311 best :1st_place_medal:

PR #206, opened 2021-01-09
Qencoder compatibility patch

natis1:
Add ffmpeg splitting, add --quiet to stop tqdm from loading, make scenedetect and tqdm not strictly required, add setup-minimal for av1an-minimal, a version with much fewer deps, when using --quiet av1an exposes the number of frames that have been encoded

### DIFF
--- a/av1an/manager/Counter.py
+++ b/av1an/manager/Counter.py
@@ -1,5 +1,10 @@
 from multiprocessing.managers import BaseManager
-from tqdm import tqdm
+has_tqdm = 1
+try:
+    from tqdm import tqdm
+except ImportError:
+    has_tqdm = 0
+
 
 def Manager():
     """
@@ -19,7 +24,7 @@ class Counter:
         self.initial = initial
         self.left = total - initial
         self.current = 0
-        self.use_tqdm = use_tqdm
+        self.use_tqdm = (use_tqdm and has_tqdm)
         if use_tqdm:
             self.tqdm_bar = tqdm(total=self.left, initial=0, dynamic_ncols=True, unit="fr", leave=True, smoothing=0.01)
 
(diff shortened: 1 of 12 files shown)


av1an/manager/Counter.py
```diff
@@ -1,5 +1,10 @@
 from multiprocessing.managers import BaseManager
-from tqdm import tqdm
+has_tqdm = 1
```
master-of-zen, line 2: yeah, this can be prettier

setup-minimal.py
```diff
@@ -0,0 +1,34 @@
+import setuptools
+
+REQUIRES = [
+    'numpy',
+    'psutil',
+    'scipy',
+    'matplotlib',
```
master-of-zen, line 7: do we really need matplotlib?

av1an/scenedetection/pyscene.py
```diff
@@ -3,10 +3,14 @@
...
-from scenedetect.video_manager import VideoManager
-from scenedetect.frame_timecode import FrameTimecode
+scenedetect = 1
+try:
+    from scenedetect.detectors import ContentDetector
+    from scenedetect.scene_manager import SceneManager
+    from scenedetect.video_manager import VideoManager
+    from scenedetect.frame_timecode import FrameTimecode
+except ImportError:
+    scenedetect = 0
```
n9Mtq4, line 13: use True/False here too

av1an/scenedetection/aom_kf.py
```diff
@@ -8,7 +8,11 @@
 from subprocess import PIPE, STDOUT
 
 import cv2
-from tqdm import tqdm
+has_tqdm = 1
+try:
+    from tqdm import tqdm
+except ImportError:
+    has_tqdm = 0
```
n9Mtq4, line 15: True/False

natis1:
Actually using true/false is suboptimal. the recommended way according to pycharm is to set the thing you would be importing to a None type, which I am doing now.

n9Mtq4 (approved):

master-of-zen:
@natis1 :1st_place_medal:

PR #207, opened 2021-01-09
Support vmaf 2.0

n9Mtq4:
Fixes vmaf validation error: vmaf 2.0 has some minimum video resolution that caused validation to fail on working configurations.


n9Mtq4:
Done in 287f213918b30a7e32937306a6c890655825729d

PR #208, opened 2021-01-09
Zero pad first pass file with 1 chunk

n9Mtq4:
Fixes an issue with first pass reuse on 1 chunk.


master-of-zen:
@n9Mtq4 :1st_place_medal:

PR #213, opened 2021-01-19
Docker

luigi311:
This will add a docker image that is built via github actions and published to the repo owners dockerhub account under the name av1an. This will also allow users to build their own via 
```
docker build -t "av1an" .
```
You will need to create a dockerhub account and a token and put them in the repository secrets settings. This still needs documentation to be added and hopefully others can suggest edits to make the image smaller as it is currently 6GB


master-of-zen:
@luigi311 🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇

PR #223, opened 2021-01-31
Docker readme and cleanup

luigi311:
This will update the dockerfile to remove the need of sudo and add basic readme information on how to use the docker container   
Also fixes PR always failing due to docker action failing. Docker action will now only run on pushes to master


master-of-zen:
@luigi311 Best, as usual ) :1st_place_medal:

PR #225, opened 2021-02-06
Update readme, use johnvansickle ffmpeg build

luigi311:
Update readme to fix word splitting on docker command

Format readme using remark

Use johnvansickles ffmpeg build, fixes issues with ffmpeg crashing causing all encodes to fail


master-of-zen:
@luigi311 🥇

PR #226, opened 2021-02-07
Changed default args on SVTAV1

Felixkruemel:
SVT-AV1's Docu only specifies using -q instead of --qp. This change was introduced with the introduction of crf based encoding instead of fixed qp.

Changed the value in the default args to match that.


master-of-zen:
@Felixkruemel 👍

PR #228, opened 2021-02-10
Docker: Use luigi311/encoders-docker, Update readme, update actions

luigi311:
Docker:
 - The docker image will now be based on luigi311/encoders-docker where all the encoders and other tools are built so the docker build github action doesnt take forever.

README:
 - Updated docker readme information to include :latest tag and --rm to prevent stale images
 - Added descriptions for all the available tags 

Actions:
 - Cleanup dependencies since the base image contains almost everything
 - Add vs chunking test
 - Use python setup.py install as pip install -r requirements was broken
 - Cat av1an log for success and failures
 - Will now tag with latest, master, #.## and sha-###### to allow users to backtrack to a previous version if necessary

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,10 +85,17 @@ jobs:
           tree -a
 
   docker:
-    runs-on: ubuntu-latest
-    needs: validate
+    runs-on: ubuntu-18.04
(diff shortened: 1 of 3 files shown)


master-of-zen:
@luigi311 :1st_place_medal:

Issue #222, opened 2021-01-26
[feature request] Vapoursynth script as source


gersch07:
First of all, thanks a lot for this great work!

One thing I'd find useful would be if a Vapoursynth script could be loaded instead of a source video file. Currently I have to transcode my Vapoursynth scripts to intermediate FFV1 MKVs using ffmpeg, and these files take up a lot of space.
In case the option to load vpy scripts already exists, a short note in the documentation would be very helpful.

lastrosade:
This could help with #221 by using imwri

adworacz:
I added support for this months ago, and have been using it personally just fine. 

I’m a little out of date from the latest commits, so if it’s not working now there’s a new bug. 

Have you tried feeding the vpy script as the input file? It should be detected and handled correctly.

adworacz:
Okay - looks like this *is* broken right now.

I believe that this is the offending line:
https://github.com/adworacz/Av1an/blob/478274a43acc8108bfd1c3433da036d1f98f5232/av1an/startup/setup.py#L88

@master-of-zen Not sure why this line was commented out. Is there a refactor going on, or was this just a temporary change that slipped through?

master-of-zen:
@adworacz It probably was a time when I was making vapoursynth by default frame selection and probing method.
My bad

adworacz:
Solid  I see you made some commits to fix - I'll give a test of the latest commit this weekend to confirm it works.

adworacz:
Looks like things aren't quite fixed yet - using my above test file, the get the following exception stacktrace:
```
WARNING: Invalid params:
'--no-sao' isn't a valid param for x265. Did you mean '--sar'?
'--deblock' isn't a valid param for x265. Did you mean '--rect'?
'-1:-1' isn't a valid param for x265. Did you mean '-1'?
'--psy-rd' isn't a valid param for x265. Did you mean '--rd'?
'--psy-rdoq' isn't a valid param for x265. Did you mean '--rd'?
0frames [00:00, ?frames/s]Output 240 frames in 0.40 seconds (602.55 fps)
Traceback (most recent call last):
  File "/home/user/src/Av1an/av1an.py", line 5, in <module>
    main()
  File "/home/user/src/Av1an/av1an/__main__.py", line 14, in main
    manager.run()
  File "/home/user/src/Av1an/av1an/manager/Manager.py", line 64, in run
    EncodingManager().encode_file(proj)
[...]

PR #232, opened 2021-02-14
Make Vapoursynth support more apparent.

adworacz:
This adds documentation in both the README and commandline arguments.

Hopefully this will prevent confusion like in #222.


master-of-zen:
@adworacz when nobody thought it could be even better ))

PR #239, opened 2021-03-25
Fix a crash when using vpy input with a space in the filename

shssoichiro:


master-of-zen:
@shssoichiro Thank you)

PR #242, opened 2021-04-05
Annotate methods for probe logging and simplified case distinction

ibhh:
Hi, 

just some minor changes that happened when I tried to fully understand your probe logging code.

P.S.: I think the log is not ideal for the fast_search results. It will print the targeted VMAF as the desired VMAF should be met by encoding with the calculated q.


master-of-zen:
@ibhh Thanks! :1st_place_medal:

PR #244, opened 2021-04-12
Kill auxiliary pipes when main encode/process is finished

Thomasedv:
Pass references to created pipes to the process_pipe functions, such that any remaining pipes can get killed. Prevents stray applications taking up RAM.

For the future, this might also make it easier to log error output from these pipes when an encoder error occurs due to issues with the piping or input files. 

Tested on Windows 10, with vspipe (lsmash I believe).


master-of-zen:
@Thomasedv Thank you :1st_place_medal:

PR #247, opened 2021-04-16
Allow folder as output for Av1an

Thomasedv:
By ending the output filename with `\` or `/`, it will be interpreted as a folder. Users will therefore be able to output encoded files to that folder. Encoded files take the standard name as if no output name were specified. The same applies to temp paths.  

Note that the temp folder will be created if it does not exist, but it will not be deleted. Each project/encode temp folder will respect the setting, but the temp folder that is specified by the user will not be deleted, in case that folder is used for other purposes, or in the event that user specifies both temp and output as the same folder. (Which is fully supported in with this pull request!)

This change forced me to change the default arg  type from Path to str, since it seems like Path objects remove the slashes at the ends of the path, which would prevent this implementation. The output file and temp attributes are saved as Path objects after processing to make sure nothing else gets broken by this change. 

If a user specifies an output folder, it will be created once a project starts, as this is the only time that folder information is present. This behavior might be worth changing to create the folder before concatenation happens. It'll require more tinkering however, whereas the current change works well without any large code changes. So I propose changing that later if that is desired.

### DIFF
--- a/av1an/arg_parse.py
+++ b/av1an/arg_parse.py
@@ -81,10 +81,10 @@ class Args:
             help="Input File or Vapoursynth (.py,.vpy) script",
         )
         io_group.add_argument(
-            "--temp", type=Path, default=None, help="Set temp folder path"
+            "--temp", type=str, default=None, help="Set temp folder path"
         )
         io_group.add_argument(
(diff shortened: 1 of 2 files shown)


master-of-zen:
@Thomasedv Thank you a lot!

PR #249, opened 2021-04-17
Use utf-8 encoding for loadscript.vpy, handles nonstandard characters.

Thomasedv:
Reported on discord, when using vapoursynth, files with uncommon characters such as ä, ö or ü.

Issue was reproduced and then fixed for me (on Windows 10, vapoursynth with lsmash) by changing loadscript.vpy encoding to utf-8. I believe python defaults to different encodings per system, and uses a different default codec causing this issue on Windows, and potentially on other OSes as well.


master-of-zen:
@Thomasedv Simple spell but quite unbreakable. Thank you)

PR #253, opened 2021-04-21
Aom firstpass fix

Thomasedv:
Added support for recent aomenc versions, which have added fields in first pass. 

Added a try/except that first tries this new mode, then defaults to old parsing, to ensure backward compatibility. 
Should fix the failing test.


master-of-zen:
@Thomasedv Thank you

PR #254, opened 2021-04-21
Determine VMAF threads value based on number of workers Closes #252

### LINKED ISSUE #252: VMAF default threads set too high
VMAF threads are currently defaulted to 12 threads per process.  However, you can expect "worker" number of VMAF processes to be spun up at any given time. (Giving you roughly worker * VMAF threads in cpu requirements)

Rather than controlling parallelism with both VMAF settings and worker settings, it should be one or the other.   Ideally (IMO) all processes would target using a ~1 core so workers can be set appropriately.

With that in mind, VMAF should be set to 1 or 2 threads by default.  Setting it higher than that will over provision a system where workers are at or close to the number of cores on a system.

cogman:
Based on discussion of #252 , Compute and set the VMAF threads based on OS CPUs to try and avoid over provisioning.

### DIFF
--- a/av1an/target_quality/target_quality.py
+++ b/av1an/target_quality/target_quality.py
@@ -1,4 +1,5 @@
 import subprocess
+import os
 
 from math import isnan
 import numpy as np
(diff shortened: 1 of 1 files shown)


master-of-zen:
@cogman Thanks, looks good for me

PR #256, opened 2021-05-08
Fix None error when user has vpoursynth, but not ffms2 or lsmash

Thomasedv:
Defaults to hybrid method in that case.


master-of-zen:
@Thomasedv :1st_place_medal:

PR #259, opened 2021-05-14
Making aom+SVT-AV1 target_quality faster and SVT-AV1 fixes

BlueSwordM:
So, these are the small changes I've been working on. 

aomenc: Removing ` "--enable-rect-tx=0"` fixes some crashes, and adding some new options speeds up the quality analysis in target_quality mode without impacting VMAF analysis scores in both live-action and animation.

SVT-AV1: Fixing the command parsing for CRF in target_quality mode, as well as adding `--keyint 240` for better efficiency and speed over the default 33 frame keyframe limit, both in normal default encoding mode and target_quality. Also, reenabling HME since it actually speeds up the encoder a bit. Since CRF is now used, tpl-la has to be kept enabled as well :smile: 

Other: probing_rate fix by making sure that all probing_rates below 4 are active. Before, 3 would go up to 4. Now, it just works.

### DIFF
--- a/av1an/encoder/svtav1.py
+++ b/av1an/encoder/svtav1.py
@@ -13,7 +13,7 @@ class SvtAv1(Encoder):
         super(SvtAv1, self).__init__(
             encoder_bin="SvtAv1EncApp",
             encoder_help="SvtAv1EncApp --help",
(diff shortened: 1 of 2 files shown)


master-of-zen:
Looks good to me @BlueSwordM :1st_place_medal: great job :+1:

PR #263, opened 2021-05-16
Use cargo workspaces, port some more vapoursynth functions to rust

redzic:
This PR uses cargo workspaces to more neatly separate the application logic and the pyo3 bindings, which will also make it much easier to remove the python FFI code at the end, and for making other frontends for the core logic, like a web interface, GUI, CLI, etc.

When there is no more python code, these vapoursynth functions should use the [vapoursynth](https://crates.io/crates/vapoursynth) crate. The reason this PR does not do that already is because, at least on my machine, it mysteriously segfaults under seemingly any circumstance, but only when being used from pyo3, not when being called by other rust code without pyo3 involved anywhere. I could go into more detail of the behavior of the segfault here, but it isn't really relevant for this PR. I'm not sure if this is an issue with pyo3 or what, but maybe an issue should be opened in their repo eventually about it. But for av1an, we should just port those functions using the vapoursynth C api later.

The rust module was renamed to `av1an_pyo3` to make it clearer that the python code is calling rust code.
I also added `vapoursynth` as a python dependency requirement. I just could not get it to work in the virtual env without it for some reason, but this should probably be removed eventually. Later on, I think we should add a flag to the build system to optionally enable vapoursynth support, which can be done using [cargo features](https://doc.rust-lang.org/cargo/reference/features.html).

There's also some extra code in there that will be properly implemented later, like the `EncodeConfig` struct, a function to directly concatenate ivf files, and some code copied from the vspipe example in the vapoursynth crate for proper linkage to the vapoursynth C api.


luigi311:
Why is the concat.rs file a thing if everything is commented out. Is that something that was uploaded by mistake?

luigi311:
Naming the rust section av1an_py03 is a good idea since right now its harder to tell what is in rust and what isn't.

redzic:
> Why is the concat.rs file a thing if everything is commented out. Is that something that was uploaded by mistake?
It was not a mistake, originally I planned on just having that in there so I didn't forget to actually integrate it later since initially the purpose of this PR was to just add the cargo workspaces, but we could also actually add a `--output-ivf` flag or something like that to actually use this function.

redzic:
Also, since the `Cargo.toml` that's used for the bindings for the pyo3 code has moved, you have to build the rust module with `maturin develop --release -m av1an-pyo3/Cargo.toml` if you're in the top-level folder.

luigi311:
You will need to update the github actions and the dockerfile then with the new build command or else its going to failed all the tests and the dockerfile too.

master-of-zen:
not sure why it say that there is merge conflict

master-of-zen:
@redzic A:clap:M:clap:A:clap:Z:clap:I:clap:N:clap:G
:1st_place_medal:

PR #264, opened 2021-05-16
Add --output_ivf flag to remux VP9 and AV1 to ivf

redzic:
Adds `--output_ivf` flag which remuxes ivf files into a single ivf file. On some inputs there still seems to be an issue with the video length in the metadata which I tried to fix by using the sum of duration in the metadata of the individual clips, but in those cases the video still plays fine, and seeking works.


master-of-zen:
@redzic Great) :medal_sports:

redzic:
looks like `vspipe` is failing in CI

master-of-zen:
:thinking:

redzic:
ok so the problem is with lsmash, it's saying it can't construct the index

redzic:
so issue is somewhere in rust implementation of `create_vs_file`, probably somewhere with resolving paths but I will have to check

redzic:
@master-of-zen should be fixed now, issue was that it was using a relative path in the vapooursynth input file for the source

master-of-zen:
@redzic :+1:

PR #269, opened 2021-05-24
Docker improvements

luigi311:
Adds docker build to all CI to test docker building

Adds pip install wheel to remove silent error with vapoursynth


master-of-zen:
@luigi311 :+1:

PR #273, opened 2021-06-05
Fix aom_keyframes, docker-publish needs validate

luigi311:
Fix typo with aom_keyframes
Docker-publish needs validate so if the tests do not pass the docker image is not published to dockerhub


master-of-zen:
@luigi311 :1st_place_medal:

PR #277, opened 2021-06-08
Probe slow

luigi311:
Fixes issues with probe_rate always being set to 4 instead of using the user supplied value.

Adds a --probe-slow flag that will use the video encoding settings for the probes. This provides the biggest benefit for x265 as it starts tuning visuals the lower the preset you go causing crfs to differ on probes than what we are shooting for.

Testing with 
https://drive.google.com/file/d/1rc6zj630oV1v5oTSh_Cj9pdvxws2BM--/view 
./av1an.py -i ~/Downloads/1080p_30s.mkv --probing-rate 2 --target-quality 94 --vmaf -e x265 -v " -p slower --crf 25 " --extra-split 60

Original
11M	1080p_30s_x265.mkv
208.7s
1080p_30s_x265_plot

Probe Slow:
6.6M	1080p_30s_x265_probe_slow.mkv
342.6s
1080p_30s_x265_plot_probe_slow


luigi311:
target quality 90
Original:
5.4M	1080p_30s_x265.mkv
179.0s
1080p_30s_x265_plot_90_original

Probe Slow:
3.5M	1080p_30s_x265.mkv
300.6s
1080p_30s_x265_plot_90_original

av1an-cli/src/lib.rs
```diff
@@ -167,5 +171,5 @@ pub fn parse_args() -> String {
 
 /// Get default values of args
 pub fn default_args() -> String {
-  serde_json::to_string(&Args::parse_from(["av1an"])).unwrap()
+  serde_json::to_string(&Args::parse_from(&["av1an"])).unwrap()
```
master-of-zen, line 174: this should be fine as it is you used nightly rust tool-chain (as recommended for pyo3)

luigi311: Should i return it to how it was then?

master-of-zen: I think it's fine

master-of-zen:
@luigi311 looks good to me :+1: 
Great job :1st_place_medal:

PR #280, opened 2021-06-14
Port logging to Rust, fix clippy warnings and remove dead code

redzic:
Calling `log` if the logging file does not exist is now a no-op. `set_log` must first be called to ensure that the file exists.

Another minor changes is that `hash_path` now encodes the hash in hex like it did in the original Python implementation, rather than formatting as an integer. Also removes `#![allow(unused)]` from the crate root and fixes the `clippy` warnings.


master-of-zen:
@redzic :1st_place_medal:

PR #282, opened 2021-06-17
Create into_vec! macro that implicitly calls `.into()` for each element

redzic:
This removes the boilerplate from before, and can potentially save some unnecessary allocations in the future. Currently, it actually adds extra allocations because `pyo3` requires cannot pass a `Cow<'_, str>` to Python, only the owned type `String`, so we convert all the elements to `String` in the Python wrapper of this function. However, if being called by other Rust code, this is unnecessary.


master-of-zen:
@redzic Great job :1st_place_medal:

PR #284, opened 2021-07-03
Port tqdm to Rust with indicatif crate

redzic:
Ports the previous tqdm code to call a progress bar implemented in Rust using indicatif instead. To achieve an fps display without flickering, a patched indicatif is needed. This should be changed to use upstream indicatif when [this issue](https://github.com/mitsuhiko/indicatif/issues/299) in indicatif is fixed.

The PySceneDetect progress bar is not affected by this PR, because it seems to manage its own progress bar.


master-of-zen:
@redzic Great work! :1st_place_medal:

PR #292, opened 2021-07-12
Add av-scenechange as the new and default scene detector

redzic:
Also fixes some clippy warnings and adds `clippy::needless_pass_by_value` since that seems to be a common code smell here.

This new scene detector also works on vapoursynth input, and may be faster in many cases. On my laptop on 720p, av-scenechange is about 46% faster than pyscenedetect.

When compiling with `-C target-cpu=native` (which on my machine enables AVX2), the perf diff is up to 53% for the same clip. The performance can be further improved by compiling av1an with PGO.

### DIFF
--- a/av1an-cli/src/lib.rs
+++ b/av1an-cli/src/lib.rs
@@ -53,8 +53,8 @@ pub struct Args {
   scenes: Option<PathBuf>,
 
   /// Specify splitting method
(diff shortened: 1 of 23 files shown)


master-of-zen:
@redzic Amazing work :1st_place_medal:

PR #295, opened 2021-07-13
Port `Project` class to Rust, use vapoursynth C api instead of vspipe for `frame_probe_vspipe`

redzic:
Also ports the suggestions (in the case of invalid params) to Rust.

Removes the batch encoding for now, because its implementation was incorrect.


master-of-zen:
Aaaaamazing :1st_place_medal:

PR #297, opened 2021-07-14
Fix chunking/resume, port chunk generation to Rust, remove unused Python code, update indicatif

redzic:
- Removed no longer used scenedetect and psutil dependencies.
- Now that indicatif has the ability to customize template keys, a fork is no longer required. Update to latest git master of indicatif.
- Port `Chunk` class and all of its methods to Rust.
- Fix broken resuming, where inconsistencies existed between `done.json` and `scenes.json`


master-of-zen:
@redzic Great work

PR #303, opened 2021-07-18
Fix workflow, Fix Dockerfile

luigi311:
- Fix workflow to run with all rust build
- Fix dockerfile with all rust build


master-of-zen:
@luigi311 That is fast) :1st_place_medal:

PR #305, opened 2021-07-19
Fix exact comparison of floats

shssoichiro:
This is a hard error in clippy,
so currently if compiled with clippy enabled,
compilation will fail. This commit allows compilation
with clippy enabled.


shssoichiro:
(The reason I casted to u64 is because these variables are already floored/ceiled right before, so this is faster than comparing to an epsilon)

PR #306, opened 2021-07-19
Fix default audio parameters

redzic:
By default, ffmpeg re-encodes the audio to vorbis if no parameters are specified.


master-of-zen:
@redzic :+1:

PR #308, opened 2021-07-19
Update indicatif branch name

shssoichiro:
They changed the name of their master branch to "main",
so this fails to build if you don't have an existing Cargo.lock file.


master-of-zen:
@shssoichiro :+1:

PR #309, opened 2021-07-19
Fix some clippy warnings, use enums instead of strings

redzic:
The `Project` struct now uses enums for `Encoder`, `SplitMethod`, and `ChunkMethod`, which is much faster than comparing up strings at runtime. This also removes many now useless functions that were wrappers around methods in `Encoder` itself which took a string as an argument.


master-of-zen:
@redzic :+1:

PR #310, opened 2021-07-19
Rename libvpx to vpx

redzic:
Previously, `-e vpx` didn't work because the `possible_values` in clap didn't contain "vpx", but vpx in `Encoder` was serialized to "libvpx", making it impossible to actually use it from the command line. This PR fixes that, while also making the naming more consistent, as it is in line with "libaom" being called just aom in the CLI.


master-of-zen:
@redzic looks good

PR #314, opened 2021-07-25
Better logging, integrate av1an-pyo3 into av1an-core

redzic:
Implements more robust logging via flexi_logger. This replaces the old
logging implementation, and is also more ergonomic, since macros from the
`log` crate are now used instead of `format!` directly. Now, warnings
and errors are also logged to stderr, in addition to the log file.

This also removes av1an-pyo3, and integrates the code into av1an-core.

Concatenation with FFmpeg is now handled on Windows differently than
other platforms through conditional compilation. FFmpeg seems to need
double backslashes in the concat file, so this is explicitly handled
in the Windows build now. Also, the spinner is now disabled on Windows
builds since the default command prompt cannot display the characters
correctly.


master-of-zen:
@redzic Aaaaaamazing

PR #316, opened 2021-07-26
Refactor: fix chunk restarting, progress bar, and potential race condition

redzic:
`DoneJson` is now a concurrent data structure, via `DashMap` and atomics
in the standard library. This should help avoid any potential race
conditions involved with one thread reading `done.json` while another
thread writes to it. Now, we only write to `done.json`, so any failure
from parsing the json while encoding is impossible.

The audio encoding is now done on a separate thread, and it is not
assumed that the audio encoding was finished if the `--resume` flag was
specified, which is now kept track of within `done.json`.

The initialization of the progress bar(s) now does not occur if the
corresponding initialization functions were not called, and the
functions to update the progress bar(s) are a NO-OP if it was not
initialized, rather than implicitly initializing it through the use of
the `Lazy` type from `once_cell`.

Additionally, all 3 concatenation functions now reside in the `concat` module.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,6 +171,7 @@ dependencies = [
  "clap 2.33.3",
  "crossbeam-channel",
  "crossbeam-utils",
+ "dashmap",
  "failure",
  "flexi_logger",
  "indicatif",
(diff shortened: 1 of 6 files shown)


master-of-zen:
@redzic amazing 🥇

PR #318, opened 2021-07-28
Fix scenes.json, update done.json when audio is done

redzic:


master-of-zen:
@redzic Best :100: :1st_place_medal:

PR #319, opened 2021-07-29
Fix target quality and VMAF plotting on Windows

redzic:
FFmpeg requires a very specific syntax to escape the file path in
Windows when using a filter, so this has been implemented with
conditional compilation. The behavior is the same on Windows, except
that the path is converted to an absolute path. Also, the only usage of
`std::env::current_dir` has been removed, since it was being used as an
ad-hoc way to convert a relative path to an absolute one, which doesn't
work if the temporary directory is in a different directory than the
current one.

We no longer kill the child processes from `process_pipe`,
since calling `wait` or `wait_with_output` closes the stdin handle
anyway.

Also, all usages of `cfg!(windows)` have been changed to `cfg!(target_os
= "windows")` to keep a consistent style.

### DIFF
--- a/av1an-core/src/encoder.rs
+++ b/av1an-core/src/encoder.rs
@@ -87,7 +87,11 @@ impl Encoder {
         into_vec![
           format!("--fpf={}.log", fpf),
           "-o",
-          if cfg!(windows) { "nul" } else { "/dev/null" },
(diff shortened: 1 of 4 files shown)


master-of-zen:
@redzic Great :1st_place_medal:

PR #335, opened 2021-08-19
Add a fast scenecut method option

shssoichiro:
Selectable with `--split-method av-scenechange-fast`.
The fast method enables downscaling the video before scene detection,
and uses av-scenechange's fast mode, which is similar to pyscenedetect.
This also removes the downscaling from the normal scenecut mode,
since downscaling does affect accuracy of the scene detection.


master-of-zen:
@shssoichiro Thanks 🥇

PR #336, opened 2021-08-19
Integrate av1an-scene-detection into av1an-core, fix some clippy warnings, update dependencies

redzic:
Also removes some vapoursynth code that was unused (which were silenced with `#[allow(unused)]`).

av-ivf and indicatif now have new releases which we are now using, so we no longer have any git dependencies either.

The reasoning for integrating av1an-scene-detection into av1an-core is that av1an-scene-detection had only one small module, which does not necessitate its own entire crate.


master-of-zen:
@redzic looks great! :medal_sports:

PR #339, opened 2021-08-20
ffmpeg: fix frame count parsing

ishitatsuyuki:
The old code used to search for the first match that looked like a frame
count. When ffmpeg emits a warning, it would also emit the frame count at
the point of warning, therefore the previous implementation gave wrong
frame count when it happened.

Fix this by modifying the code to take the last match. Also replace the
index of cap by an explicit one since this is the index of capture group
(which is constant given the regex), not matches.


ishitatsuyuki:
The frame count is still not correct actually. I will investigate this.

master-of-zen:
@ishitatsuyuki Thanks) Looking up for it

ishitatsuyuki:
OK, I think this can be merged as-is. The frame count isn't correct if the input has timestamp gaps (a VFR-ish thing), but that should be worked on along with other VFR issues.

master-of-zen:
@ishitatsuyuki Thanks :+1:

PR #341, opened 2021-08-21
Replace some panics with Err returns

ishitatsuyuki:
Errors now look like this:

```
Error: Input file "foo" does not exist!
```

instead of this:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Input file "foo" does not exist!', av1an-cli/src/main.rs:116:27
```


master-of-zen:
@ishitatsuyuki Thanks, looks great :1st_place_medal:

PR #345, opened 2021-08-28
Lazily compile all regexes, move macros to util module

redzic:
This should avoid some unnecessary overhead where regexes were being recompiled every time their respective function was called.


master-of-zen:
@redzic looks good :1st_place_medal:

PR #347, opened 2021-08-29
Remove other unused dependencies

redzic:
It seems like these dependencies don't need to be specified to compile, as they are already specified in the root Cargo.toml.


master-of-zen:
👍

PR #349, opened 2021-08-31
Add output file confirmation prompt, improve error handling slightly

redzic:
This is something the Python implementation did but removed and never added back until now.


master-of-zen:
Looks good :+1:

PR #351, opened 2021-09-02
Refactoring, use ffms2 or lsmash for frame counting on chunks if avai…

redzic:
…lable

- This removes the `parking_lot::Mutex` wrapping the multi progress bar, as it was not actually necessary. This also removes `parking_lot` as a dependency, as it is unused now.

- We now evaluate a vapoursynth script in-memory (i.e. without creating any temporary files) to get the frame count for chunks if available. Note that ffms2 is preferred over lsmash for getting the frame count, as ffms2 works with `--enable-keyframe-filtering=2` in aomenc, but lsmash does not.

- Rework the implementation of `mkvmerge` concatenation to not perform any unnecessary UTF-8 validation or allocations. This requires an algorithm to specify chunks with a '+' in between, but without a trailing '+', without later removing arguments (as the API of `std::process:Command` does not support this).

- Also fix some miscellaneous clippy warnings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -200,7 +200,6 @@ dependencies = [
  "log",
  "num_cpus",
  "once_cell",
- "parking_lot",
  "path_abs",
  "plotters",
  "regex",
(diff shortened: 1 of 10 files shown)


redzic:
looks like the version of  ffms2 in the CI doesn't support AV1, hence why it is working for x264/5 and vpx, but not for aom, rav1e, and SVT-AV1

redzic:
Until we figure out the issues with ffms2/lsmash (they do not work in the CI for some reason, and lsmash will actually be worse than ffmpeg for getting the frame count with clips encoded with `--enable-keyframe-filtering=2` as it just crashes), we cannot change the behavior of the frame probes in this way. I will open another PR with the other changes.

PR #353, opened 2021-09-03
Prevent data loss if concatenation failed

mxsrm:
I had multiple cases in which concatenation using `mkvmerge` silently failed and av1an deleted the temp folder afterwards. I saw the `mkvmerge` process running in the background but it never wrote to the output file.

Of course the real fix should include further debugging and location of the problem with mkvmerge, but this patch at least prevents data loss. People can either join the files manually or simply run av1an with `--resume` without losing possible hundreds of hours of encoding time.

### DIFF
--- a/av1an-core/src/lib.rs
+++ b/av1an-core/src/lib.rs
@@ -1273,7 +1273,9 @@ impl Project {
         plot_vmaf(&input, &output_file, model).unwrap();
       }
 
-      if !keep {
+      if !Path::new(&output_file).exists() {
+          warn!("Concatenating failed for unknown reasons! Temp folder will not be deleted: {}", temp);
+      } else if !keep {
         if let Err(e) = fs::remove_dir_all(temp) {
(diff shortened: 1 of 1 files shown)


redzic (commented):
I think it would be clearer to write it something like this:
```
if !Path::new(&output_file).exists() {
  // warn concatenation failed
} else if !keep {
  // delete temp folder
}
```

mxsrm:
I could debug the problem and it is a mkvtoolnix bug; nevertheless I think having this kind of failsafe is good to have.

Sometimes mkvmerge detects hevc bitstream files as the wrong format, uses the wrong multiplexer and ofc cant append them:

```
File '.623664b/encode/00479.mkv': container: DTS
Track ID 0: audio (DTS-ES)
```

I reported the bug to upstream and hopefully it will soon be fixed.

redzic (approved):
LGTM

master-of-zen:
@mxsrm Thanks! :+1:

PR #354, opened 2021-09-04
Refactor: Remove `parking_lot::Mutex` from multi progress bar, rework…

redzic:
… mkvmerge concatenation implementation

- This removes the `parking_lot::Mutex` wrapping the multi progress bar, as it was not actually necessary. This also removes `parking_lot` as a dependency, as it is unused now.

- Rework the implementation of `mkvmerge` concatenation to not perform any unnecessary UTF-8 validation or allocations. This requires an algorithm to specify chunks with a '+' in between, but without a trailing '+', without later removing arguments (as the API of `std::process:Command` does not support this). This does not change the behavior of `-c mkvmerge` in any way, however.

- Also fix some miscellaneous clippy warnings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -200,7 +200,6 @@ dependencies = [
  "log",
  "num_cpus",
  "once_cell",
- "parking_lot",
  "path_abs",
  "plotters",
  "regex",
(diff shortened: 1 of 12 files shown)


master-of-zen:
@redzic Looks good :+1:

PR #355, opened 2021-09-07
Print scenechange info to stderr when in verbose mode

shssoichiro:
I found myself frequently having to look in the log file to see this,
and felt it would be more convenient to have it printed to the terminal.
This only affects verbose mode--normal mode will continue to print
this info to only the log file.


master-of-zen:
@shssoichiro Thanks :1st_place_medal:

PR #356, opened 2021-09-07
Fix printing on verbose scenechange info

shssoichiro:
Because of the `scenes` variable updating,
this would print the number of scenes after
extra splitting twice instead of the number of scenes
before splitting and then the number after splitting.


master-of-zen:
@shssoichiro :+1:

PR #361, opened 2021-09-17
Fix confirmation prompt

### LINKED ISSUE #359: Output file confirmation prompt reads input with "\r" on Windows
As stated in the title, whenever you press enter during the prompt, av1an will interpret the enter keypress as "\r", thus making the code fail and go through an infinite loop.
```
Output file ".\\output.mkv" exists. Do you want to overwrite it? [Y/n]: n
Sorry, response "n\r" is not understood.
Output file ".\\output.mkv" exists. Do you want to overwrite it? [Y/n]: y
Sorry, response "y\r" is not understood.
Output file ".\\output.mkv" exists. Do you want to overwrite it? [Y/n]:
Sorry, response "\r" is not understood.
```

redzic:
Fixes #359 and also is more clear that pressing enter will not overwrite.

### DIFF
--- a/av1an-cli/src/lib.rs
+++ b/av1an-cli/src/lib.rs
@@ -181,14 +181,12 @@ fn confirm(prompt: &str) -> io::Result<bool> {
     stdout.flush()?;
     stdin.read_line(&mut buf)?;
 
-    match buf.as_str() {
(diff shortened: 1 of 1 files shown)


master-of-zen:
@redzic looks good :+1:

PR #362, opened 2021-09-20
Routine dependency update

shssoichiro:


master-of-zen:
@shssoichiro Wouldn't hurt :)

PR #363, opened 2021-09-27
Tweaked default settings for better target-quality+normal encoding

BlueSwordM:
These changes are rather minor, but actually quite important in some contexts.

1. `--kf-max-dist=240`. By default, aomenc and libvpx-vp9 have non adaptive keyframe intervals of 9999 frames. With av1an's default maximum chunk length of 240, it does not pose a problem. 

However, if a user wanted to extend that chunk length for some reason(higher target-quality accuracy by preventing early skips or less chunks to encode), defaulting the maximum keyframe interval to 240 would allow the video streams to keep good seekability. Therefore, setting the max KF interval to 240 for libvpx-vp9 and aomenc ois sound.

1.1. `--kf-min-dist=12`. For some reason, aomenc and libvpx-vp9 have a default minimum keyframe interval of 0. This is not a good encoder decision since any kind of flash in the video stream might force the encoder to put a keyframe in with nothing to reference to, cranking up the bitrate for no reason for a few frames ahead. This also makes target-quality behave badly in edge cases, since flashes are taken into account of the score, and the increase in bitrate makes VMAF perceptually undershoot the quality. It is not large, but it it not negligible either.

As such, a default of `--kf-min-dist=12` allows for slightly higher encoding efficiency in normal encoding mode since it forces the placement of non-keyframe golden frames, and higher accuracy for target-quality.

2. For the VP9 changes, they are pretty simple. `--cpu-used=2` in place of `--cpu-used=0` makes for a 30-50% faster encoder for a barely noticeable metric loss. `--auto-alt-ref=6` is a profile=2 feature, and increases the efficiency by 5-7% over a standard `--auto-alt-ref=1` stream by virtue of a more complex frame-hierarchy.


shssoichiro:
I actually think this is unnecessary, we should have keyframe detection disabled in the actual encoder. This is because all keyframe detection is done ahead of time by av1an's scenechange code (based on rav1e), so each encoder receives exactly one segment which should have exactly one keyframe, at the very start.

BlueSwordM:
@shssoichiro That is true, but only if the scene-detection was absolutely perfect and scene flashes that went by undetected weren't an issue.

As it stands, that is currently not the case for many sources.

shssoichiro:
I wonder if this will be the case after https://github.com/xiph/rav1e/pull/2813 is merged? My hope is that it will not be.

The alternative is to disable aomenc's scene detection so it doesn't override av1an's decisions.

redzic:
looks like the vpx params are not correct, the encoder gives the error:
```
Failed to initialize encoder: Invalid parameter
            kf_min_dist not supported in auto mode, use 0 or kf_max_dist instead.
```

redzic:
@BlueSwordM I think I agree with @shssoichiro in that ideally we would disable scene detection within the encoder itself as has been done in rav1e with `--no-scene-detection`. However, I think adding `--kf-min-dist=12` is probably fine for now.

As far as preventing early skips, I am not exactly sure why adding extra splits would decrease target quality accuracy, since ultimately the keyframe placement would be the same as just letting the encoder add the extra splits with `--kf-max-dist`. Is it just that VMAF works better on slightly longer continuous clips since it is more temporally aware of what is happening in the scene? If that is the case, I would actually prefer if av1an could distinguish between extra splits and original keyframes, and make target quality decisions based on original keyframes, but still force keyframes at extra splits.

BlueSwordM:
@redzic 

1. Disable 2-pass in aomenc is not a good idea, as it does things other than scene-detection, so unless the aom devs change that behavior, we'll never be able to disable 2-pass.

2. It's basically that having longer chunks within a same scene might let VMAF know there's a lower quality scene in that chunk, meaning it could increase the bitrate(by decreasing the Q/CRF).

3. In that regard, I'll also update the PR with something to fix aomenc in the meantime because of this nasty bug:
https://bugs.chromium.org/p/aomedia/issues/detail?id=3154

PR #368, opened 2021-09-30
Use a different method for ffmpeg frame count

### LINKED ISSUE #367: Encoding fails with "FRAME MISMATCH" when encoding with aomenc --enable-keyframe-filtering=2
This is likely due to the fact that keyframe filtering mode 2 is not supported by all decoders. Whatever decoder av1an is using is failing to decode the files.

I'll self-assign this and take a look.

shssoichiro:
This method uses ffprobe to count the number of packets
(which is identical to the number of frames, but faster)
in a video stream. This works with more video formats,
including with --enable-keyframe-filtering=2 in aomenc.
Performance should be similar or better than ffmpeg -copy.

Fixes #367

### DIFF
--- a/av1an-core/Cargo.toml
+++ b/av1an-core/Cargo.toml
@@ -44,6 +44,10 @@ path_abs = "0.5.1"
 av-scenechange = "0.7.0"
 y4m = "0.7.0"
 
+[dependencies.ffmpeg-next]
+version = "4.4.0"
+features = ["static"]
+
 [dependencies.vapoursynth]
 version = "0.3.0"
 features = ["vsscript-functions", "vapoursynth-functions"]
--- a/av1an-core/Cargo.toml
+++ b/av1an-core/Cargo.toml
@@ -46,7 +46,6 @@ y4m = "0.7.0"
 
 [dependencies.ffmpeg-next]
 version = "4.4.0"
-features = ["static"]
 
 [dependencies.vapoursynth]
 version = "0.3.0"
@@ -59,3 +58,6 @@ features = ["rt", "process", "io-util"]
 [dependencies.dashmap]
 version = "4.0.2"
 features = ["serde"]
+
+[features]
+ffmpeg_static = ["ffmpeg-next/static", "ffmpeg-next/build"]
--- a/av1an-core/src/project.rs
+++ b/av1an-core/src/project.rs
@@ -91,6 +91,8 @@ pub struct Project {
 impl Project {
   /// Initialize logging routines and create temporary directories
   pub fn initialize(&mut self) -> anyhow::Result<()> {
+    ffmpeg_next::init()?;
+
     info!("File hash: {}", hash_path(&self.input));
 
     self.resume = self.resume && Path::new(&self.temp).join("done.json").exists();
--- a/av1an-core/src/project.rs
+++ b/av1an-core/src/project.rs
@@ -387,6 +387,18 @@ impl Project {
       self.video_params = self.encoder.get_default_arguments();
     }
 
+    if self.encoder == Encoder::aom
+      && self.concat != ConcatMethod::MKVMerge
+      && self
+        .video_params
+        .iter()
+        .any(|param| param == "--enable-keyframe-filtering=2")
+    {
+      bail!(
+        "keyframe filtering mode 2 currently only works when using mkvmerge as the concat method"
+      );
+    }
+
     self.validate_input();
     self.initialize().unwrap();
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -90,6 +90,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: check
+          args: --no-default-features
 
       - name: Build Av1an
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ on:
       - "*.md"
 
 env:
-  deps: tree
+  deps: tree llvm clang
   DEBIAN_FRONTEND: noninteractive
 
 jobs:
@@ -82,14 +82,14 @@ jobs:
             default: true
             components: rustfmt, clippy
 
+      - name: Install requirements
+        run: |
+          apt-get update && apt-get install -y ${{ env.deps }}
+
       - name: Run cargo check
         uses: actions-rs/cargo@v1
         with:
           command: check
-
-      - name: Install requirements
-        run: |
-          apt-get update && apt-get install -y ${{ env.deps }}
 
       - name: Build Av1an
         run: |
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,7 +147,7 @@ dependencies = [
  "av-data",
  "av-format",
  "log",
- "nom",
+ "nom 6.1.2",
 ]
 
 [[package]]
@@ -174,7 +174,7 @@ dependencies = [
  "path_abs",
  "serde",
  "serde_json",
- "shlex",
+ "shlex 1.1.0",
  "structopt",
 ]
 
@@ -187,7 +187,7 @@ dependencies = [
  "ctrlc",
  "path_abs",
  "serde",
- "shlex",
+ "shlex 1.1.0",
  "structopt",
 ]
 
@@ -203,6 +203,7 @@ dependencies = [
  "crossbeam-channel",
  "crossbeam-utils",
  "dashmap",
+ "ffmpeg-next",
  "flexi_logger",
  "indicatif",
  "itertools",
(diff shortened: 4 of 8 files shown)


shssoichiro:
I originally wrote that VS was still broken. That's wrong because av1an uses ffms2 instead of lsmash. I manually tested with lsmash because it was easier. 🤦‍♂️

shssoichiro:
I want to finish through a full encode just to verify nothing else needs changed. 10min left.

redzic:
Just putting this out there, but alternatively we can even statically link ffmpeg (using the `ffmpeg-next` crate) into av1an and get the frame count that way.

Here is an example program that does that (using `ffmpeg-next` 4.4.0):

```rust
use ffmpeg::format::input;
use ffmpeg::media::Type;
use std::env;

fn main() -> Result<(), ffmpeg::Error> {
    ffmpeg::init().unwrap();

    if let Ok(mut ictx) = input(&env::args().nth(1).expect("Cannot open file.")) {
        let input = ictx
            .streams()
            .best(Type::Video)
            .ok_or(ffmpeg::Error::StreamNotFound)?;
        let video_stream_index = input.index();

        let frames = ictx
            .packets()
            .filter(|(stream, _)| stream.index() == video_stream_index)
            .count();

        println!("{} frames", frames);
    }

    Ok(())
}
```

shssoichiro:
@redzic I don't mind that. For the most part I prefer it to calling a program on the system, however it would increase compile times. So that might be worth a bigger discussion if we want to pull that crate in.

shssoichiro:
Findings:

- ffmpeg concat method doesn't work at all. Resulting file is 500 bytes, doesn't play.
- ivf concat method produces a video file, but seeking is broken.
- mkvmerge concat method works and produces a playable, seekable file.

shssoichiro:
I feel like getting ffmpeg's concat method to work with kf=2 will be harder than parsing the frame count. So for now I added a validation check to exit at the start of the encode if they're setting kf=2 and not using mkvmerge concat method.

av1an-core/src/project.rs
```diff
@@ -387,6 +387,17 @@ impl Project {
       self.video_params = self.encoder.get_default_arguments();
     }
 
+    if self
+      .video_params
+      .iter()
+      .any(|param| param == "--enable-keyframe-filtering=2")
```
redzic, line 393: I think we should skip this step if not using aomenc as the encoder

av1an-core/src/project.rs
```diff
@@ -387,6 +387,18 @@ impl Project {
       self.video_params = self.encoder.get_default_arguments();
     }
 
+    if self.encoder == Encoder::aom
+      && self
+        .video_params
+        .iter()
+        .any(|param| param == "--enable-keyframe-filtering=2")
+      && self.concat != ConcatMethod::MKVMerge
```
redzic, line 395: one more nitpick, `&&` evaluates left to right, so the check for mkvmerge should be before the param check for enable-kf=2 just to avoid comparing strings if it's not necessary

redzic:
LGTM, although I remember @master-of-zen said he would prefer to use only the ffmpeg binary and not ffprobe when I mentioned this issue before with `--enable-keyframe-filtering=2`.

@master-of-zen how would you feel about pulling in ffmpeg-next?

master-of-zen:
@shssoichiro @redzic Thank you both.
I don't like adding ffprobe as another dependency that is not integrated into av1an. That's the primary reason why I gave favor to FFmpeg frame counting.
I feel totally fine with adding `ffmpeg-next`, as it's statically linked, and I don't need to expect from end-user to have the correct version of it that builds correctly.
And I like the idea of integrating into av1an instead of relaying on third-party packages, which can vary tremendously.
I'm not adamant about gatekeeping ffprobe, I would like to continue without adding it.
That's how I see it

shssoichiro:
I'm fine with going the ffmpeg-next route. Follow-up question: Do we want to try and replace all calls to ffmpeg with the ffmpeg-next crate? I think that would help as it cuts out that entire external dependency.

master-of-zen:
@shssoichiro That would be aaaaaamazing :)

redzic:
@shssoichiro Ideally we would, but for this PR you can change just `num_frames`. We can incrementally port the rest of the calls to ffmpeg.

shssoichiro:
Cool, I can do that as a followup PR then. Since I have nothing better to do right now 😅

av1an-core/Cargo.toml
```diff
@@ -44,6 +44,10 @@ path_abs = "0.5.1"
 av-scenechange = "0.7.0"
 y4m = "0.7.0"
 
+[dependencies.ffmpeg-next]
```
redzic, line 47: Can we use ffmpeg-next but refer to it as ffmpeg by changing the package name in cargo? Specify the dependency as ffmpeg but add a property `package = "ffmpeg-next"`

shssoichiro: I attempted that, and it was easier to just use `ffmpeg_next` because we have a module called `ffmpeg` in av1an and that becomes confusing to both devs and the compiler.

shssoichiro:
Oh... That broke CI. 😬

av1an-core/Cargo.toml
redzic: ah I see, makes sense. what do you think of calling it `avcodec` then, or do you prefer `ffmpeg_next`?

shssoichiro: I'm not a fan of using `avcodec` since technically that's just one part of ffmpeg. I also don't love how `ffmpeg_next` is kind of verbose, but I'm not sure of a great way to shorten it. 🤔

redzic: I know there's another crate called `ffav` (but then, it might be confused since we are not actually using that package). But for now, I guess we can just keep it as `ffmpeg_next`

shssoichiro:
I had to enable the "build" feature in ffmpeg to get the CI to pass... I'm going to try a couple more things to see if I can get it to work without that, because that does slow down the builds considerably.

redzic:
For some reason the without the `build` feature, ffmpeg doesn't seem to actually get linked statically.

Some comparisons of binary size of the program to get the frame count posted earlier:


|              | with debug symbols | stripped  |
|--------------|--------------------|-----------|
| build+static | 104.56 MB          | 20.96 MB  |
| static       | 3.44 MB            | 294.91 KB |

Maybe we could add a cargo feature to control if ffmpeg gets linked dynamically or statically, and link it dynamically in CI?

shssoichiro:
Which would it make more sense to use as the default, static or dynamic?

redzic:
Maybe we should statically link it by default, since the entire point of pulling this crate in is to remove runtime dependencies. I think we would mostly use static linking for releases, and dynamic linking for doing local builds and for the CI.

shssoichiro:
It looks like there's no capability in Cargo to set features per profile (e.g. automatically use static linking for release builds). That would have been nice. Given that, I wonder if it still makes sense to statically link, because that would be enabled for dev builds by default? (We can easily control CI with CLI args, so I'm not too concerned about that, it's more about developer + user experience.)

redzic:
From a user experience, it would be easiest if they could just do `cargo build --release` and it statically links everything by default. Also, while we're at it, can we set only `av-scenechange` to compile with maximum optimizations for debug builds?

And from a developer experience, it would be easier for it to dynamically link by default. On my machine for the example program posted earlier, incremental static builds took 2.37s and incremental dynamic builds took 0.74s.

> It looks like there's no capability in Cargo to set features per profile (e.g. automatically use static linking for release builds)

That is unfortunate, because I feel like the perfect solution if cargo could do that would have been to set `av-scenechange` to compile with opt-level=3 for both debug and release builds, and statically link on release builds by default and dynamically link on debug builds.

shssoichiro:
We can set opt-level on a per-crate and per-profile basis, so that is possible for av-scenechange. However, profiles do not support changing features, so automatically switching between dynamic and static is not possible as far as I'm aware.

shssoichiro:
I decided to make static the default, and I disabled it for CI builds (so they will use dynamic). I think that's the better option for compatibility, given the goal was to remove an external dependency. Devs can always use the dynamic build with `cargo build --no-default-features`.

shssoichiro:
I remembered why the CI was originally failing. The version of ffmpeg in the docker image is too old and can't decode AV1 (or at least that's what it seems). So we will have to use the static version in CI, it seems.

shssoichiro:
This is ready for any final feedback now.

redzic:
LGTM, @master-of-zen ?

master-of-zen:
@shssoichiro @redzic Amazing :rocket:

redzic:
Ah, the only thing actually was that the commit message is out of date now, it still says that it uses ffprobe

PR #371, opened 2021-10-03
Allow disabling extra splits by setting to 0

shssoichiro:


master-of-zen:
@shssoichiro I don't know how it slipped through)
Thanks 👍 🚀

PR #372, opened 2021-10-04
Rewriting simpler ffmpeg calls using the API

shssoichiro:
These functions were the easiest to convert. Transcoding with the API is significantly more complicated, and will be a work-in-progress.


master-of-zen:
@shssoichiro :1st_place_medal: :rocket:

PR #374, opened 2021-10-04
Refactor into `Input` type for handling vapoursynth input

redzic:
Fixes several issues, namely `--vmaf` not working on vapoursynth input
as well as unifying the VMAF interface, which also results in
`--vmaf-threads` being passed correctly to `--vmaf`, which is massively
faster now as a result. If `--vmaf-threads` is not specified, then
the number of detected CPU threads is used for `--vmaf`.

Additionally, VMAF plots are rendered as SVGs now, which also removes
some dependencies since we don't use the bitmap backend at all anymore.

Bunch of other small fixes as well:

- Check if mkvmerge is installed if `--concat mkvmerge` was specified
- Do not implicitly use the default arguments if parsing the video/audio/ffmpeg args failed
- Better error messages if ffmpeg is missing
- Update to sysinfo 0.20.0 as mingw now has rust 1.55
- Just use `String` instead of `VecDeque` for storing encoder output
- Delete unused `file_validation.rs`
- Move remaining VMAF structs to `vmaf.rs`
- Many other small code style changes.


shssoichiro:
Code looks fine to me, my only suggestion is in the future to split these out into smaller PRs. Having several unrelated changes in one commit makes it harder to review and removes the ability to bisect in case something breaks.

redzic:
Yeah, sorry I have been working on this on and off for a while now so it's all in one massive PR.

master-of-zen:
@redzic Great work!

PR #375, opened 2021-10-05
Fix many clippy warnings, many small code style changes

redzic:
Mostly just stylistic changes/cleanups, but some notable changes:

- Validating encoder params now operates over string views of the original help text output rather than allocating new strings for everything
- The `Project` struct has been renamed to `EncodeArgs`, and the `project` module has been renamed to `settings`


master-of-zen:
@redzic Great! :1st_place_medal: :rocket:

PR #376, opened 2021-10-08
Set log level in FFmpeg + small fixes

redzic:
Also replaces trivial regexes with standard library functions as suggested by clippy.


master-of-zen:
@redzic Good job! 🚀

PR #378, opened 2021-10-12
Do not split encoder args for non-vpx-based CLIs

shssoichiro:
This fixes an issue with encoders such as x264
where parsing would fail if you passed a negative
argument, such as `--deblock -1:-1`.


master-of-zen:
@shssoichiro Bellissimo :ok_hand:

PR #379, opened 2021-10-13
Reduce piping overhead to encoder in certain situations

redzic:
Previously, we always piped the chunk method into FFmpeg
to apply filters and convert the pixel format, but this
just adds unnecessary overhead if there is no FFmpeg filter
specified and if the pixel format of the source and encode
are the same. So, in that situation, we now pipe the chunk
method directly into the encoder instead. This should
nicely speed things up a bit (and possibly reduce memory
consumption).

Additionally, `--pix-format` is now verified using the FFmpeg
API when the CLI args are parsed, making it impossible to
accidentally specify an incorrect format. Previously, there
were no checks at all, so if you specified the wrong format it
would just fail during the encoding process with no obvious
indication of the root cause.

This does not apply to vapoursynth input for now, as that
would be slightly more complicated since it has a different
API for getting the pixel format.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,8 +179,8 @@ dependencies = [
  "anyhow",
  "av1an-core",
  "ctrlc",
+ "ffmpeg-next",
  "path_abs",
- "serde",
  "shlex 1.1.0",
  "structopt",
 ]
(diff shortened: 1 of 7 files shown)


master-of-zen:
@redzic nice 🚀

PR #381, opened 2021-10-13
Changes for rav1e's hybrid scene detection update

shssoichiro:
This pulls in the changes from https://github.com/xiph/rav1e/pull/2813 into av1an. Combined with https://github.com/xiph/rav1e/pull/2827, the new method should be as fast as or faster than the previous slow method, while being more accurate.

Currently this is pointed at rav1e's git master awaiting a new release from them. We can either leave this PR open until a release is made, or merge it and take out the reference to rav1e's git when they perform a release. Up to the maintainers here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,8 +147,7 @@ dependencies = [
 [[package]]
 name = "av-scenechange"
 version = "0.7.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


master-of-zen:
@shssoichiro Very nice :rocket: :1st_place_medal:

redzic:
I also wanted to mention that this pulls in xiph/rav1e#2819, which makes the fast mode of scene detection even faster.

PR #382, opened 2021-10-14
Fix skipping of ffmpeg pipe step

shssoichiro:
This optimization added in #379 was never triggered.
Because the `initialize()` method always added a set
of default params to `self.ffmpeg_params`, the clause
to skip the ffmpeg pipe, `self.ffmpeg_params.is_empty()`,
was unreachable.


master-of-zen:
@shssoichiro only issue, doesn't work with target quality

redzic:
Perhaps `compose_ffmpeg_pipe` should always add those parameters and take in the pixel format, so that it isn't left up to the caller to make sure it is correct?

shssoichiro:
Combined into #383, since this was required in order to be able to test those changes.

PR #383, opened 2021-10-16
Scene detect changes

shssoichiro:
Unfortunately, this is a large PR, because many of these changes are either required for the code to work or made refactoring much easier.

- No longer convert input to 8-bit for scene detect

It appears that downconverting does have an impact on scene detection accuracy. The speed penalty for running in 10-bit compared to 8-bit is only 10%. Given this, it seems preferable to run the scene detect in the more accurate format. This also means that, for Vapoursynth input, we can skip the extra ffmpeg pipe if scenecut downscaling is not being used.

- Fix conditional use of newer Vapoursynth APIs

These were never working for users who tried to enable the newer Vapoursynth APIs. This also enables the new APIs by default, because the vast majority of users should have a recent enough version of Vapoursynth to use the 3.1 vsscript API. (If they're not, I don't know how any of their plugins are still working.)

- Turn off the ffmpeg build feature from defaults

In addition to the issues reported by Windows users as a result of this feature, libaom reports `None` for the pixel format of a video, making it unusable for detecting the bit depth of the input if input is AV1. I attempted to get the ffmpeg-next build of libdav1d working, but ran into issues,  I believe because Arch's package manager only includes a dynamic library for libdav1d and not a static version. We should leave this feature disabled by default until we can get it to a more stable level.

- Fix skipping of ffmpeg pipe step

This optimization added in #379 was never triggered. Because the `initialize()` method always added a set of default params to `self.ffmpeg_params`, the clause to skip the ffmpeg pipe, `self.ffmpeg_params.is_empty()`, was unreachable.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,7 +94,7 @@ jobs:
 
       - name: Build Av1an
         run: |
-          cargo build --release
+          cargo build --release --features ffmpeg_static
 
       - name: Download videos
         run: |


shssoichiro:
It turns out I was misled. `zscale` is actually significantly _slower_ than `scale`, so I will not be pursuing that option.

The change to re-add `--sc-downscale-height` is up.

master-of-zen:
@shssoichiro Amazing work :1st_place_medal: :+1:

PR #396, opened 2021-11-01
Clean up installation instructions and add crates.io badge

svenstaro:
I just put av1an into Arch Linux's [community] repository so this should be reflected in the README.


master-of-zen:
@svenstaro  Nice, thank you :medal_military: :rocket:

PR #400, opened 2021-11-06
Always copy subtitles and all audio tracks

redzic:
- Require mkvmerge for x265, since we were actually concatenating x265 with ffmpeg by secretly creating an mp4 even if the file extension was mkv, and mp4 does not support certain types of subtitles that many sources have. ffmpeg cannot concatenate raw hevc bitstreams without the timestamps properly set without first creating an intermediate mp4 file, so for this reason we require mkvmerge to be used when using x265.

- Document behavior of how to encode separate audio tracks differently with this change, as using `-map` will not work and will probably cause unexpected results. All audio tracks and subtitles (technically, everything that isn't a video track) are copied by default now.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,7 +105,7 @@ jobs:
 
       - name: Testing ${{ matrix.name }}
         run: |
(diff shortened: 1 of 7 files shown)


n00mkrad:
Does this account for container compatibility, as WEBM doesn't support most audio/subtitle codecs? 
I think it would be best to limit this feature to MKV outputs.

master-of-zen:
@redzic best, as usual :wink: :rocket:

PR #401, opened 2021-11-11
Initial round of improvements for error handling

redzic:
The main change that this commit introduces is an `EncoderCrash` type
for errors, which also reports the stderr of the process piping to
the encoder so that it is easier to tell what the actual error is.

Target quality probe crashes are now properly logged with the
`EncoderCrash` type, replacing the old panicking code and making the
error significantly easier to read.

An "optimization" has been implemented which does not wait for the child
process piping to the encoder to exit, and instead collects its stderr
output asynchronously and immediately exits upon an encoder crash. This
does not yet apply to target quality probes, as there is not yet a
single interface for running an encoder (it is done manually for now).
This will change in the future, but this patch is getting too big for
everything so for now we just capture stderr synchronously for target
quality probes.

- Also update `get_percentile` to use an O(N) algorithm from the
  standard library instead of O(N*log N).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,12 +208,14 @@ dependencies = [
  "log",
  "num_cpus",
  "once_cell",
+ "parking_lot",
  "paste",
  "path_abs",
  "plotters",
  "regex",
  "serde",
  "serde_json",
+ "simdutf8",
  "splines",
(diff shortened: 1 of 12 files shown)


master-of-zen:
@redzic amazing

PR #402, opened 2021-11-15
Allow setting thread affinity for each worker

redzic:
Introduces a new command line option called `--set-thread-affinity=n`,
where `n` specifies the number of threads to assign to each worker.

We use the `affinity` crate instead of the `nix` crate directly to
get abstractions for Windows and Linux.


master-of-zen:
@redzic amazing, big POG :rocket: :1st_place_medal:

PR #403, opened 2021-11-16
Fix compilation on macos

redzic:
Disabled setting thread affinity on platforms other than linux and windows. This should fix compilation on macos.


master-of-zen:
@redzic nice :rocket:

PR #404, opened 2021-11-16
Improve logging (especially with `--verbose`)

redzic:
Much more information is now logged when using `--verbose` (all INFO
messages), and the output is also much cleaner.

The issue of logs being hidden while the progress bar is active
because of conflicting ANSI escape codes has been completely fixed, and
no more undesirable hacks are required to correctly display the
set_thread_affinity warning message correctly, for example. This has
been achieved by using indicatif's `println` function instead of just
printing to stderr directly.

flexi_logger has been moved to av1an-cli as it should be, since a
logging framework that actually prints things and handles log level
filtering is not the responsibility of the core library. However, the
actual progress bar code with indicatif still lives in av1an-core for
now.


redzic:
The failing test case seems to be unrelated to the code changes in this PR, and rather to do with the docker pull failing.

master-of-zen:
@redzic amazing, as usually :rocket: :man_dancing:

PR #409, opened 2021-11-17
Fix a case where the log file could fail to be created

shssoichiro:
Previously, these warn statements came before we create
the temp directory. I'm assuming when the first log
statement is run, flexi_logger then attempts to create
the log file. If it fails, it doesn't try again.

Ensuring we create the temp directory before attempting
to log should resolve this issue.


redzic:
@shssoichiro amazing :rocket:

master-of-zen:
@shssoichiro Pog indeed :100:

PR #418, opened 2021-11-19
Enable thin-LTO for release builds

shssoichiro:
Although the majority of av1an's time is spent
in third-party binaries, compiling with thin-LTO
does show up to a 10% improvement on scene detection
time. This will cause compilation to be slower.
However, thin-LTO has far less of a compilation time
impact than full LTO, while providing almost the same
speedups.


redzic:
@shssoichiro :+1:

PR #428, opened 2021-11-22
Min chunk size to 24 and faster more accurate aomenc target-quality

BlueSwordM:
With this PR, we'd change the minimum scene length to 24 frames as to make it more psycho-visually accurate if there are rapid scene-changes.

As for the 2nd part, disabling all kinds of temporal filtering and denoising in aomenc is a good idea for target-quality in itself, since we need the highest image possible out of a certain Q to evaluate quality properly, even if it creates more artifacts.


redzic:
@BlueSwordM amazing :rocket: :rocket: :rocket: :rocket:

PR #431, opened 2021-11-25
Fix all clippy warnings as of latest stable rustc

shssoichiro:
I kind of wish that we wouldn't use clippy::pedantic and
clippy::nursery, as the number of `allow` lints at the top of src/lib.rs
is becoming quite absurd. However, I at least wanted to clear out the
large set of warnings being issued, regardless of that discussion.

### DIFF
--- a/av1an-cli/src/lib.rs
+++ b/av1an-cli/src/lib.rs
@@ -300,7 +300,7 @@ fn confirm(prompt: &str) -> io::Result<bool> {
 }
 
 pub fn parse_cli(args: CliOpts) -> anyhow::Result<EncodeArgs> {
-  let temp = if let Some(path) = args.temp {
+  let temp = if let Some(path) = args.temp.as_ref() {
     path.to_str().unwrap().to_owned()
   } else {
     format!(".{}", hash_path(args.input.as_path()))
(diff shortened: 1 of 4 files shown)


av1an-cli/src/lib.rs
```diff
@@ -300,7 +300,7 @@ fn confirm(prompt: &str) -> io::Result<bool> {
 }
 
 pub fn parse_cli(args: CliOpts) -> anyhow::Result<EncodeArgs> {
-  let temp = if let Some(path) = args.temp {
+  let temp = if let Some(path) = args.temp.as_ref() {
```
redzic, line 303: Personally I prefer `as_deref()` here, since it returns an `Option<&Path>` instead of a `Option<&PathBuf>`. Is there some other reason to change it to use `as_ref()`? It doesn't seem that `as_deref` causes any clippy warnings, at least from the ones we have enabled now

shssoichiro: Mainly, for consistency. We had some `as_ref` and some `as_deref`. My personal preference is `as_ref` but I don't have a real reason for that, just habit, I suppose. The performance should be equivalent regardless of the type difference.

redzic: OK, that's fine. I guess the use of `as_ref` and `as_deref` are pretty much the same here anyway (especially since it looks like `as_deref` just explicitly calls `as_ref` and `deref`), and I guess it could make the code slightly more confusing to use both, especially if you're reading it outside of an IDE with type annotations.

PR #438, opened 2021-12-09
Show scenecuts detected even in non-verbose mode

shssoichiro:


redzic:
I think I would prefer to still have the `info!` to do this, since that actually adds it to the log file. Also don't we already sort of display this with the `Queue:`?

shssoichiro:
We sort of display it, but only the number after extra cuts are applied.

In the end, I don't really care too much as I realize this is a personal preference. So I'll let this one go.

PR #439, opened 2021-12-09
Add kbps and final file size estimates to progress bar

shssoichiro:
This looks at the size of each chunk when it completes and saves that
size in kilobytes into the done.json file. For backwards compatibility
for encodes resumed from an earlier version of av1an, if the done.json
does not contain the chunk size, then on resume, av1an will lookup the
size of that chunk. The estimates on the progress bar also handle
resumes and chunk crashes gracefully.

### DIFF
--- a/av1an-core/src/lib.rs
+++ b/av1an-core/src/lib.rs
@@ -112,6 +112,14 @@ impl Input {
       Input::VapourSynth(path) => vapoursynth::num_frames(path.as_path()).expect(FAIL_MSG),
     }
   }
+
+  pub fn frame_rate(&self) -> f64 {
+    const FAIL_MSG: &str = "Failed to get frame rate for input video";
+    match &self {
+      Input::Video(path) => ffmpeg::frame_rate(path.as_path()).expect(FAIL_MSG),
+      Input::VapourSynth(path) => vapoursynth::frame_rate(path.as_path()).expect(FAIL_MSG),
+    }
+  }
 }
 
 impl<P: AsRef<Path> + Into<PathBuf>> From<P> for Input {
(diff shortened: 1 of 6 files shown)


av1an-core/src/lib.rs
```diff
@@ -112,6 +112,14 @@ impl Input {
       Input::VapourSynth(path) => vapoursynth::num_frames(path.as_path()).expect(FAIL_MSG),
     }
   }
+
+  pub fn frame_rate(&self) -> f64 {
+    const FAIL_MSG: &str = "Failed to get number of frames for input video";
```
redzic, line 117: Probably just a copy/paste error, but this fail message should probably say failed to get the framerate, instead of number of frames

redzic:
Would it be possible to get the framerate of the input once at the beginning and cache it for future use?

shssoichiro:
Yeah, the only time it pulls frame rate an extra time is if it's resuming. So a maximum of two times.

av1an-core/src/lib.rs
```diff
@@ -129,11 +137,19 @@ impl<P: AsRef<Path> + Into<PathBuf>> From<P> for Input {
   }
 }
 
+#[derive(Debug, Deserialize, Serialize, Clone, Copy)]
+struct DoneChunk {
+  frames: usize,
+  // Currently an Option for backwards compatibility,
+  // since this field was added recently.
+  size_kb: Option<u32>,
```
redzic, line 145: I don't think this would actually work to maintain backwards compatibility, as serde serializes `Option::None` in json in this case as `size_kb: null`, so the old way of serializing it would not work. I also just tried it on an older temp folder, and this is the output I got:

```
Error: invalid type: integer `147`, expected struct DoneChunk at line 1 column 34
```

Theoretically we still *could* maintain backwards compatibility by trying to parse the json both ways, but I think that gets ugly pretty fast. If we are going to switch away from json eventually anyway, I think it would be ok to just drop backwards compatibility for this, and perhaps we could land this in a 'nightly' or 'experimental' branch for now as discussed in discord before. What do you think?

av1an-core/src/progress_bar.rs
```diff
@@ -153,10 +161,41 @@ pub fn dec_mp_bar(dec: u64) {
   }
 }
 
+pub fn update_mp_bar_info(kbps: f64, est_mb: f64) {
+  if let Some((_, pbs)) = MULTI_PROGRESS_BAR.get() {
+    pbs
+      .last()
+      .unwrap()
+      .set_message(format!(", {:.1} Kbps, est. {:.1} MB", kbps, est_mb));
```
redzic, line 169: indicatif seems to have a wrapper for displaying bytes in a more human readable way: https://docs.rs/indicatif/0.16.2/indicatif/struct.HumanBytes.html
Perhaps we could just use that (or something similar) instead of just always displaying it in megabytes?

redzic:
landing this in the new `unstable` branch as this contains breaking changes

PR #440, opened 2021-12-09
Use bincode instead of serde_json

redzic:
This would go into unstable as this definitely breaks backwards compatibility.

A binary serialization format should have much less overhead compared to json.


shssoichiro:
It seems like this is breaking all of the target-quality tests. :thinking:

PR #444, opened 2021-12-14
Retry chunk without TPL if a TPL-related crash occurs

shssoichiro:
aomenc has had a history of crashes related to TPL on certain
chunks, particularly in videos with less motion, such as animated
content. This commit enables a workaround to retry a chunk with TPL
disabled if such a crash is detected. Although there is some amount of
psychovisual quality loss with TPL disabled, this is preferable to being
unable to complete the encode.

### DIFF
--- a/av1an-core/src/broker.rs
+++ b/av1an-core/src/broker.rs
@@ -2,7 +2,7 @@ use crate::{
   ffmpeg, finish_multi_progress_bar, finish_progress_bar, get_done,
   progress_bar::{dec_bar, dec_mp_bar},
   settings::EncodeArgs,
-  Chunk, Instant, TargetQuality, Verbosity,
+  Chunk, Encoder, Instant, TargetQuality, Verbosity,
 };
 use std::{
   fmt::{Debug, Display},
@@ -14,6 +14,7 @@ use std::{
 };
 
 use cfg_if::cfg_if;
+use memchr::memmem;
 use smallvec::SmallVec;
 use thiserror::Error;
 
@@ -61,6 +62,15 @@ impl From<Vec<u8>> for StringOrBytes {
 impl From<String> for StringOrBytes {
   fn from(s: String) -> Self {
     Self::String(s)
+  }
+}
+
+impl StringOrBytes {
+  pub fn as_bytes(&self) -> &[u8] {
+    match self {
+      Self::String(s) => s.as_bytes(),
+      Self::Bytes(b) => &b,
+    }
   }
 }
 
@@ -173,9 +183,12 @@ impl<'a> Broker<'a> {
     }
 
     // Run all passes for this chunk
+    let mut tpl_crash_workaround = false;
     for current_pass in 1..=self.project.passes {
       for r#try in 1..=self.max_tries {
-        let res = self.project.create_pipes(chunk, current_pass, worker_id);
+        let res = self
+          .project
+          .create_pipes(chunk, current_pass, worker_id, tpl_crash_workaround);
         if let Err((e, frames)) = res {
           if self.project.verbosity == Verbosity::Normal {
             dec_bar(frames);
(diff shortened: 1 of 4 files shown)


master-of-zen:
Looks nice to me)

av1an-core/src/broker.rs
```diff
@@ -193,6 +196,11 @@ impl<'a> Broker<'a> {
           // avoids double-print of the error message as both a WARN and ERROR,
           // since `Broker::encoding_loop` will print the error message as well
           warn!("Encoder failed (on chunk {}):\n{}", chunk.index, e);
+
+          if format!("{:?}", e.stderr).contains("av1_tpl_stats_ready") {
```
redzic, line 200: Can we do this without an allocation and UTF-8 checking? It looks like we already have `memchr` in our dependency tree transitively, so we could use something like [`memmem`](https://docs.rs/memchr/2.4.1/memchr/memmem/index.html) to just search for a byte slice directly, without having to do another allocation or verify the UTF-8.

redzic:
Also, just generally, if we have any encoder-specific workarounds we should check that we are encoding with the buggy encoder that needs the workaround before checking if the workaround is needed.

redzic:
Also it would be nice if we could have a comment explaining why the workaround is needed/that aomenc is buggy

av1an-core/src/broker.rs
shssoichiro: I'd actually like to figure out how to do that, if someone has worked with av1an's `StringOrBytes` type before? Because that is what we are working with here. Is the thing to do just a `match` and test both branches?

redzic: Yeah actually now that you mention that the type is `StringOrBytes`, this wouldn't actually work for the `Bytes` case since "Aa" would be formatted as `"raw bytes: [65, 97]"` (since it's using debug formatting, see `Debug` impl for `StringOrBytes`), so you can't actually check the string directly. But I think pattern matching and using `.contains()` would be fine in the string case, and using `memmem` in the bytes case.

redzic: Actually, we can just always use `memmem` since it should work for the regular string types also. You'd just have to call `.as_bytes()` on the string. That's probably the simplest way to do it.

redzic: Actually, I think the absolute simplest and cleanest way to do it would to just add an `as_bytes` function in the impl of `StringOrBytes` itself so that there is no code duplication in the future either. Then we'd just call newly-added `as_bytes` method and pass the bytes to `memmem`.

av1an-core/src/broker.rs
```diff
@@ -193,6 +206,20 @@ impl<'a> Broker<'a> {
           // avoids double-print of the error message as both a WARN and ERROR,
           // since `Broker::encoding_loop` will print the error message as well
           warn!("Encoder failed (on chunk {}):\n{}", chunk.index, e);
+
+          if self.project.encoder == Encoder::aom
+            && memmem::rfind_iter(e.stderr.as_bytes(), b"av1_tpl_stats_ready")
```
redzic, line 211: Very small and pedantic nitpick but it appears that it may be slightly simpler to just do `memmem::rfind(..).is_some()` rather than `rfind_iter`. It also appears that `rfind` as opposed to `rfind_iter` uses the Rabin-Karp algorithm for small (<64 bytes) haystacks, otherwise uses the fallback impl.

av1an-core/src/settings.rs
```diff
@@ -186,25 +186,26 @@ impl EncodeArgs {
...
     worker_id: usize,
+    tpl_crash_workaround: bool,
   ) -> Result<(), (EncoderCrash, u64)> {
     let fpf_file = Path::new(&chunk.temp)
       .join("split")
       .join(format!("{}_fpf", chunk.name()));
 
+    let mut video_params = self.video_params.clone();
+    if tpl_crash_workaround {
+      video_params.push("--enable-tpl-model=0".to_string());
```
redzic, line 197: If the user explicitly enables `--enable-tpl-model=1`, does this conflict with that? I'm not familiar how the arg parsing in aomenc works.

shssoichiro: I just gave it a test and it looks like whatever is specified last will override earlier instances of the arg. Which is what we want here.

redzic:
LGTM

PR #445, opened 2021-12-14
Add mergify config

shssoichiro:
This config includes the following rules:

- Automatically merge once a PR passes all tests and is approved.
- Automatically backport each PR from our main branch into our stable branch


shssoichiro:
As mentioned on discord, we will need to finalize what we want our branching strategy to look like. The bot, and git in general, seem to work best when master is the unstable branch, and there is a stable branch which is backported to. Our current setup is the opposite, the downside of that is that rebasing to a head branch is really, really discouraged in git, so porting from master to unstable would result in the commits from master being in front of the unstable commits. The downside there, of course, is it becomes harder to find which commits are on unstable but not on master.

redzic:
While we're at it, can we make the CI run the unit tests in av1an-core via `cargo test`?

redzic:
Another option is to leave these unstable features behind a feature flag within cargo (kind of like what rav1e does), but some of this would require rewriting the arg parsing to not use structopt as I believe there is no way to use conditional compilation from structopt, but it is possible from clap directly.

redzic:
I think I prefer having a different branch though, as it allows us to be more flexible in what kinds of changes we can make. We could make large refactors while breaking backwards compatibility, for example.

shssoichiro:
I agree, I think it is easier having separate branches. If it is within the same branch, larger changes like migrating from structopt to clap will require a lot of hacks for backwards compatibility. Even for smaller changes, it becomes annoying to maintain two versions of the same code. I don't think av1an is such a large or system-critical project that we want to deal with that.

redzic:
If the bot automatically backports every PR to stable, then to me it doesn't make sense two have two separate branches. How would we distinguish between the actually experimental changes that shouldn't be backported and the changes that should be backported?

shssoichiro:
Ah, yeah, I forgot to include that in the config file. The way their documentation recommends is that we would have some sort of Github label, e.g. "unstable", that we add to the PR and use that to indicate that it shouldn't be merged to the stable branch.

.mergify.yml
```diff
@@ -0,0 +1,61 @@
+pull_request_rules:
+  - name: automatic merge for main when CI passes and approved
+    conditions:
+      - "#approved-reviews-by>=1"
+      # Unfortunately, we cannot just say "all CI must pass".
```
redzic, line 5: Would it be possible to just consolidate all the tests into a single test so that we only need one `check-success`? I think that might make the CI itself faster too, as it could avoid recompiling av1an for each test. Perhaps we could also circumvent the need to download the test videos, and just generate very short (like <1s) 64x64 black screen videos to use as test clips.

redzic:
Some ideas to speed up the CI even more:

- Remove the `cargo check` step, since if there are any compiler errors they should be revealed by the `cargo build --release` step, so it's basically just adding an extra 3 minutes to the CI times for catching the very uncommon case of compiler errors slightly more quickly.
- Disable LTO for the CI (perhaps we could use the fancy new cargo profiles in rust 1.57 and create a profile just for the CI)
- Use dynamically linked ffmpeg (we found that this wasn't possible a while ago, but it may be worth investigating again)
- Use `cargo test --release` to avoid having to recompile in debug mode for the tests
- Use `--sc-method=fast` for everything (probably a relative minor speedup but maybe still worth it)
- Use very simple, extremely short and very low resolution videos (like I mentioned in my other comment, I believe we can just generate one with ffmpeg, like 20 frames, 64x64, black screen yuv420p should be pretty fast)

Also we should probably run all the tests with av1an's `-y` flag just in case.

redzic:
Also I'm not sure what the `tree -a` for every command was originally intended for. It seems to just be cluttering up the CI output and slowing it down a little bit. It's probably just left over from the python version, but I don't think we need it anymore.

shssoichiro:
I'd like to run at least the unit tests in debug mode, in order to catch overflows and debug assertions. I'm even considering making those turn on for the integration tests, but I'm not sure how much that would impact performance. Maybe not much if we use fast scenechange. I think the rest of the ideas make sense.

redzic:
I think it'd be fine if we run the integration tests in debug mode if we use `--sc-downscale-height=1` (I just tested this and it actually seems to work). We could also just disable scene detection, but I don't want to do that because I want the scene detection to be tested as well.

shssoichiro:
I can try it with the new profile option to see how bad performance is if we use release mode, but enable debug assertions and overflow checks. Probably not great because overflow checks are expensive, but it's worth testing, I think.

Cargo.toml
```diff
@@ -40,3 +40,9 @@ opt-level = 3
 [profile.release]
 lto = "thin"
 codegen-units = 1
+
+[profile.ci]
+inherits = "release"
```
redzic, line 45: We might also want to set `codegen-units = 16` (or even higher, 16 is the default) in our CI profile to increase parallel compilation. Also apparently `lto = false` still performs "thin local LTO" (https://doc.rust-lang.org/cargo/reference/profiles.html#lto), so we could also experiment with `lto = "off"` as well.

PR #446, opened 2021-12-14
Add extra conditional for tpl_crash_workaround

redzic:
I think this would probably be good if somehow aomenc happens to crash *again* even with the tpl workaround. Realized this after I merged #444.


shssoichiro (approved):
I think the behavior doesn't change, but it does avoid searching the stderr output again, so it's worthwhile enough.

redzic:
The only change is that it doesn't print the warning message twice I believe, as well as what you mentioned

PR #447, opened 2021-12-15
`cargo update` + revert indicatif

redzic:
Indicatif replaced its regex with a hand-written parser, but this unfortunately breaks `wide_bar` and thus our progress bar. We will have to report an issue upstream about this, but until that is reported and fixed we have to use the old regex.


redzic:
Just going to merge this as this was the old behavior

PR #451, opened 2021-12-15
Fix clippy warnings

redzic:
I disabled `clippy::needless_pass_by_value` since it's mostly catching false-positives nowadays (like complaining about a type that is technically not `Copy` but is just internally a `u64`, although it was useful in the beginning when there was a lot of passing `Vec<i32>` by value).


redzic:
@Mergifyio rebase

shssoichiro (approved):

shssoichiro:
@Mergifyio rebase

shssoichiro:
@Mergifyio rebase

PR #455, opened 2021-12-16
Update indicatif + `cargo update`

redzic:
The issue with the new parser has been fixed as of the latest commit, and the new parser is much faster than the old regex. `indicatif` now no longer depends directly on regex, but `console` still uses regex to parse ansi escape codes. Eventually it should be possible to remove regex entirely as a non-build/dev dependency.


shssoichiro (approved):
Hooray, we're getting closer 👏

PR #460, opened 2021-12-21
Automatically add cq-level arg if it is not present

shssoichiro:
Currently, target quality requires a cq-level (or equivalent for the
encoder) arg to be present in order to work, so that it can replace the
existing cq-level with the new one. This commit makes it so that if the
arg is not present, av1an will insert it with the proper quality level.

### DIFF
--- a/av1an-core/src/encoder.rs
+++ b/av1an-core/src/encoder.rs
@@ -1,4 +1,5 @@
 use crate::{ffmpeg::compose_ffmpeg_pipe, inplace_vec, into_vec, list_index};
+use arrayvec::ArrayVec;
 use cfg_if::cfg_if;
 use ffmpeg_next::format::Pixel;
 use itertools::chain;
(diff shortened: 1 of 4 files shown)


shssoichiro:
Re. arrayvec: I think it will be useful for us in other places as well, so I used it here in what seemed like a good use case (a small vec with a variable number of elements).

redzic:
I think it might be better to just remove all instances of the q args at the start, and then just append the target quality params to the modified params rather than having to find the index to replace each time. What do you think?

shssoichiro:
Yeah, I think that makes sense even if just for the reason that we only need one function to build the new args instead of two.

shssoichiro:
Actually looking at the code now, we would still need `q_match_fn` and `insert_q` which is where the string duplication is. So I don't think we really gain anything by removing and then inserting as opposed to replacing or inserting.

redzic:
Alright that's fine. Can we remove the hacks in the CI tests that specify the cq-level/crf for each encoder so that we can have this be integration tested?

av1an-core/src/encoder.rs
```diff
@@ -340,20 +341,40 @@ impl Encoder {
   fn replace_q(self, index: usize, q: usize) -> (usize, String) {
     match self {
       Self::aom | Self::vpx => (index, format!("--cq-level={}", q)),
-      Self::rav1e | Self::svt_av1 | Self::x265 | Self::x264 => (index + 1, format!("{}", q)),
+      Self::rav1e | Self::svt_av1 | Self::x265 | Self::x264 => (index + 1, q.to_string()),
```
redzic, line 344: I changed this to use `format!("{}", q)` as opposed to `q.to_string()` since `format!` appears to generate less code. See this godbolt example here: https://godbolt.org/z/d4aWsr9z3

I'm open to changing it to `to_string` as well, but as far as I understand right now I think `format!` is better (not sure why they even generate different assembly in the first place)

shssoichiro: That is very interesting that `format` generates better assembly. I would expect format to generate worse assembly. Clippy also has a lint that should trigger when doing something like this, I'm surprised it was not. That kind of seems like something that should be reported to the Rust compiler issues.

redzic (approved):

PR #462, opened 2021-12-22
Automatically set one-pass mode if --rt was specified

redzic:


av1an-core/src/settings.rs
```diff
@@ -535,6 +535,14 @@ properly into a mkv file. Specify mkvmerge as the concatenation method by settin
       );
     }
 
+    if matches!(self.encoder, Encoder::aom | Encoder::vpx)
+      && self.passes == 2
```
shssoichiro, line 539: I wonder if we want to make this `>= 2` in case some crazy person tries to use 3-pass with `--rt`?

redzic: I'll do that in case we ever support 3-pass, although av1an will not let you set anything other than 1 or 2 for passes

shssoichiro (approved):

PR #464, opened 2021-12-22
Bump av-ivf to 0.2.3

shssoichiro:
av-ivf was updated to nom 7. This removes several duplicate versions of
dependencies, reducing the total number of dependencies for av1an by 7.


redzic:
Thanks!

redzic (approved):

PR #469, opened 2021-12-24
Add more info to --help

redzic:


shssoichiro:
Just to be certain: The intent here is to add more details within the code comments, correct? When running the `-h` command, it only shows the first line from each item (i.e. prior to the first empty linebreak). I think it would be cool to have `-h` show the short version and a `--fullhelp` to show the more detailed info. I don't know if structopt allows that. clap does, but we have not done that migration yet.

redzic:
@shssoichiro No it's supposed to show up in `--help` as well

Currently, `-h` shows short output, and `--help` is basically `--fullhelp` like you describe (where these additions do show up)

shssoichiro:
That's weird. But, I guess if that's how structopt works, then we'll roll with it.

shssoichiro (approved):

PR #471, opened 2021-12-25
Regular dependency upgrades

shssoichiro:
This includes minor bumps to flexi_logger, sysinfo, vergen, and dashmap. In the past, I believe there were some issues with sysinfo that prohibited us from bumping it, but they seem to now be doing regular releases again to push out the bug fixes we needed. flexi_logger required an added feature to work around an issue with local time. We do not actually log times, but the interface in `LogWriter` requires passing `DeferredNow` which uses local time, which requires us to add this feature or have an obnoxious warning logged every time av1an starts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,9 +49,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.51"
+version = "1.0.52"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


redzic (approved):
Looks good, thanks! There was some strange issue with flexi_logger with printing some error that seems to have been fixed now as of this PR (edit: Oh I guess that's what you were referring to with the extra feature).

PR #482, opened 2021-12-31
Reduce length of mkvmerge command

redzic:
This will not fully fix mkvmerge on Windows yet but this should at least reduce the length of the executed mkvmerge command by using relative paths. A full fix that breaks up the mkvmerge command into multiple commands if there are too many chunks is coming later.


shssoichiro (approved):
The code looks okay. I'll assume you've tested it, since I don't have a windows machine with av1an to test with.

n00mkrad:
chad

PR #485, opened 2021-12-31
Add option to generate film grain tables for aomenc

### LINKED ISSUE #364: [Feature Request][aom] av1an to control the denoising and noise table generation process
aomenc's denoising sucks so I had the idea that av1an could optionally take ffmpeg denoising filter parameters. Currently the workflow is split the video into chunks, queue up and instantiate workers: each worker runs probes and then encodes the chunk. The optional new workflow would be split the video into chunks, queue up and instantiate workers: each worker will then run ffmpeg to denoise the chunk according to manual parameters, [noise-model](https://aomedia.googlesource.com/aom/+/master/examples/noise_model.c) generates a grain table with the original and denoised chunk, worker encodes the chunk. Maybe add a toggle for whether the original or denoised chunk is encoded.

Then we can combine av1an's chunked vmaf-q probing/encoding with non-crappy denoising for the best of both worlds.

A simpler solution is for the user to generate a noise table for the entire source video and pass in whatever original or denoised input they want to be encoded and av1an would split the noise table file for the correct chunks, but the table generation seems to be completely single-threaded so would really benefit from being chunked...

I don't know any Rust but I've poked through the source code so may take a stab at a PR in the future but not for a while.
Open questions include whether to apply the other video filters at the same time as denoising for the most accurate modelling, but this would also mean they would have to be run on the source file too resulting in large yuv intermediates, Although it wouldn't be so bad since the number of active workers is not infinite.

shssoichiro:
Closes #364

@redzic Do you have any thoughts on whether this should go into stable or unstable?


redzic:
Wow! As long as it doesn't break resuming/serialization, I think it can go into stable.

redzic:
This will probably take me a while to review.

shssoichiro:
> Wow! As long as it doesn't break resuming/serialization, I think it can go into stable.

It shouldn't break either of those. Even if someone resumes while using the new option, the worst that happens is their film grain will look different. If they resume without using the new option, it will look exactly the same as before.

av1an-core/src/settings.rs
```diff
@@ -952,6 +965,42 @@ properly into a mkv file. Specify mkvmerge as the concatenation method by settin
...
+      if !grain_table.exists() {
+        debug!(
+          "Generating grain synth file at ISO {}",
+          u32::from(grain) * 100
+        );
+        let (width, height) = match &self.input {
+          Input::VapourSynth(video) => crate::vapoursynth::resolution(video)?,
+          Input::Video(video) => crate::ffmpeg::resolution(video)?,
+        };
+        let transfer = match &self.input {
```
redzic, line 979: It might be worth making these methods on the `Input` type itself.

shssoichiro:
Note: The noise table generation code was ported from the tool within aomenc, and then modified to our needs.

av1an-core/src/settings.rs
```diff
@@ -526,6 +529,16 @@ properly into a mkv file. Specify mkvmerge as the concatenation method by settin
       self.video_params = self.encoder.get_default_arguments();
     }
 
+    if let Some(grain) = self.grain_synth {
+      if grain > 64 {
+        bail!("Valid strength values for grain synth are 0-64");
+      }
+    }
+
+    if self.grain_synth.is_some() && self.encoder != Encoder::aom {
```
redzic, line 538: We can probably move this check to be inside if the `if let Some(grain)` statement

av1an-core/src/grain.rs
```diff
@@ -0,0 +1,161 @@
...
+
+  // Also reasonable values for current cameras. The read noise is typically
+  // higher than this at low ISO settings but it matters less there.
+  const PHOTO_RESPONSE_NON_UNIFORMITY: f32 = 0.005;
+  const INPUT_REFERRED_READ_NOISE: f32 = 1.5;
+
+  // Focal plane exposure for a mid-tone (typically a 18% reflectance card), in lx·s.
+  let mid_tone_exposure = 10. / args.iso_setting as f32;
+
+  // Assumes a 35mm sensor (36mm × 24mm).
```
redzic, line 79: Would it be reasonable to allow a configurable sensor size for more fine-tuning?

BlueSwordM:
Holy shit.

av1an-core/src/grain.rs
BlueSwordM: Yes, as it would allow for large grains, which is one weakness of how grain synthesis normally works.

It can look wonky however, which is why normal grain synth+better grain preservation encoder side wise is better.

shssoichiro: What would this configuration look like? I'm asking because I'm certainly no expert on camera sensors. This math was done by whoever wrote the photon noise tool in aomenc's repo.

BlueSwordM: I know just the right person for this, but I don't think he'll be free for a little while.

shssoichiro: Are we talking days, or weeks? If it's more than a couple of days, I'd move to go forward with this PR for now and make an issue to follow up later to add this configuration option.

redzic: Let's tweak this later then and move forward with the PR.

BlueSwordM: Days, no real worries.

In the worst case scenario, we can just do what you said Soichiro.

redzic:
My only concerns are that 1. we do not implement transfer characteristics other than SMPTE ST 2084 and BT.1886 (and just implicitly fallback to BT. 1886 if it is neither of those), and 2. that the option is called `--grain-synth`, which might be confusing to the user.

I think we can address 1. in a later PR, and for 2. I think we could rename the option to something like `--photon-noise`, and document that e.g. you don't need to use this option to use film grain synthesis in SVT-AV1.

shssoichiro:
The reason for implementing only BT.1886 for SDR is that, according to what I could find, transfer characteristics for BT.601, BT.709, and SDR BT.2020 are all equivalent to BT.1886 when viewed on an LCD monitor. If you look at stats when playing back any of them in mpv, for example, it shows all of them as `BT.1886`. The reason for using this as a fallback is because, I wasn't quite sure how to handle all of the oddball transfer options that are rarely used in practice.

redzic (approved):
LGTM, I looked over the `photon_noise_table` example in aomenc as well and the code here seems to be equivalent.

PR #487, opened 2022-01-01
Migrate to clap-v3

redzic:
Clap v3 has a derive proc macro very similar to the structopt one (enabled via the `derive` feature), so not that much work was needed to port everything over.

Changes from structopt:
- Adds help headings to organize the `--help` better
- `-v`/`--video-params`, `-f`/`--ffmpeg`, and `-a`/`--audio-params` all no longer require leading whitespace for the argument to be recognized. This means you can do e.g. `-v "--cpu-used=10 --rt"` and it will work without issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,11 +147,8 @@ version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c5f1ee878fc434a52b8fc258418a9b1c08273c48599da652cdd79c43d6ad9f10"
 dependencies = [
- "clap",
  "rav1e",
  "rust_hawktracer",
- "serde",
- "serde_json",
  "y4m",
 ]
 
@@ -167,7 +164,6 @@ dependencies = [
  "serde",
  "serde_json",
  "shlex 1.1.0",
- "structopt",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


shssoichiro:
Did this end up breaking anything? We originally talked about this being an unstable change, but maybe it doesn't need to be?

Cargo.lock
```diff
@@ -147,7 +147,7 @@ version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c5f1ee878fc434a52b8fc258418a9b1c08273c48599da652cdd79c43d6ad9f10"
 dependencies = [
- "clap",
+ "clap 2.34.0",
```
shssoichiro, line 150: We should've done this a long time ago, but we should disable the `binary` feature in `av-scenechange` to avoid having it pull in clap 2. We can do this with just `default-features = false` in the dependency line for it.

redzic: Thanks! Disabled the default features for av-scenechange.

redzic:
> Did this end up breaking anything? We originally talked about this being an unstable change, but maybe it doesn't need to be?

In its current form, it does not break anything. So I guess this change as it is can go into stable.

shssoichiro (approved):

PR #493, opened 2022-01-02
Add a CLI option for changing chunk order

shssoichiro:
People have been requesting this because the default chunk order tends
to skew toward underestimating the final filesize. Longest first is kept
as the default because it provides the best performance.


av1an-core/src/settings.rs
```diff
@@ -578,7 +580,20 @@ properly into a mkv file. Specify mkvmerge as the concatenation method by settin
...
+        chunks.sort_unstable_by_key(|chunk| Reverse(chunk.frames));
+      }
+      ChunkOrdering::ShortestFirst => {
+        chunks.sort_unstable_by_key(|chunk| chunk.frames);
+      }
+      ChunkOrdering::Sequential => {
+        // Already in order
+      }
+      ChunkOrdering::Random => {
+        chunks.sort_unstable_by_key(|_| rand::random::<u16>());
```
redzic, line 594: Let's use [shuffle](https://docs.rs/rand/latest/rand/seq/trait.SliceRandom.html#tymethod.shuffle) for this, just to guarantee an O(n) time complexity

shssoichiro: Thanks, I was looking for this method and couldn't find it. 👍

av1an-core/src/lib.rs
```diff
@@ -232,6 +232,20 @@ pub enum ChunkMethod {
   LSMASH,
 }
 
+#[derive(
+  PartialEq, Eq, Copy, Clone, Serialize, Deserialize, Debug, Display, EnumString, IntoStaticStr,
+)]
+pub enum ChunkOrdering {
+  #[strum(serialize = "long_to_short")]
```
redzic, line 239: This is quite pedantic, but it seems more consistent to serialize these with dashes rather than underscores

redzic (approved):

PR #498, opened 2022-01-03
Automatic tiling for default settings

master-of-zen:
Adds automatic tiling for aomenc, rav1e, svt-av1, vpx.
Encoder options are added if tiling is required.
Picture without tiles represented as (1, 1) tiles.
Tiles are set calculated from resolution, separately for horizontal and vertical resolution.
Default tile is 720 pixels, and tile calculation rounds down, so 1280x720 is (1,1) tiles, 1920x1080 is (2,1), and so on.

### DIFF
--- a/av1an-core/src/encoder.rs
+++ b/av1an-core/src/encoder.rs
@@ -232,44 +232,114 @@ impl Encoder {
   }
 
   /// Returns default settings for the encoder
-  pub fn get_default_arguments(self) -> Vec<String> {
+  pub fn get_default_arguments(self, tiles: (u32, u32)) -> Vec<String> {
     match self {
       // aomenc automatically infers the correct bit depth, and thus for aomenc, not specifying
       // the bit depth is actually more accurate because if for example you specify
       // `--pix-format yuv420p`, aomenc will encode 10-bit when that is not actually the desired
       // pixel format.
-      Encoder::aom => into_vec![
-        "--threads=8",
-        "--cpu-used=6",
-        "--end-usage=q",
-        "--cq-level=30",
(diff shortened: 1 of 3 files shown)


av1an-core/src/encoder.rs
```diff
@@ -232,44 +232,118 @@ impl Encoder {
...
-      ],
+      Encoder::aom => {
+        let defaults: Vec<String> = into_vec![
+          "--threads=8",
+          "--cpu-used=6",
+          "--end-usage=q",
+          "--cq-level=30",
+        ];
+
+        match tiles.0 > 1 || tiles.1 > 1 {
```
shssoichiro, line 249: Doing `match` on a bool may be confusing to other people looking at this code. I think doing an `if ... else` would be clearer and would avoid needing to write the match arms. (same for the other instances below)

master-of-zen: ok

av1an-core/src/encoder.rs
```diff
@@ -232,44 +232,115 @@ impl Encoder {
...
+          chain!(defaults, aom_tiles).collect()
+        } else {
+          defaults
+        }
+      }
+      Encoder::rav1e => {
+        let defaults: Vec<String> =
+          into_vec!["--speed", "6", "--quantizer", "100", "--no-scene-detection"];
+
+        match tiles.0 > 1 || tiles.1 > 1 {
```
shssoichiro, line 274: Missed one 🙁

master-of-zen: kekw

shssoichiro (approved):

PR #500, opened 2022-01-03
Correct and simplify tiles calculation

master-of-zen:
Correctly using log2. convertion to u32 should floor.
Less total lines


master-of-zen:
Looks good to me

redzic (approved):

PR #507, opened 2022-01-03
Remove some unnecessary allocations

redzic:
In `encoder.rs` there are a bunch of small but unnecessary heap allocations. We could make further improvements by using `SmallVec` in other places as well, but this should be good enough for a single PR.

### DIFF
--- a/av1an-core/src/encoder.rs
+++ b/av1an-core/src/encoder.rs
@@ -1,4 +1,4 @@
-use crate::{ffmpeg::compose_ffmpeg_pipe, inplace_vec, into_vec, list_index};
+use crate::{ffmpeg::compose_ffmpeg_pipe, inplace_vec, into_array, into_vec, list_index};
 use arrayvec::ArrayVec;
 use cfg_if::cfg_if;
 use ffmpeg_next::format::Pixel;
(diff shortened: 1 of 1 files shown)


av1an-core/src/encoder.rs
```diff
@@ -36,31 +36,31 @@ impl Encoder {
...
-        into_vec!["-o", output, "-"],
+        into_array!["-o", output, "-"],
       )
       .collect(),
       Self::rav1e => chain!(
-        into_vec!["rav1e", "-", "-y"],
+        into_array!["rav1e", "-", "-y"],
         params,
-        into_vec!["--output", output]
+        into_array!["--output", output]
```
shssoichiro, line 47: I'm a little surprised this works because I thought you couldn't store a heap-allocated String in a stack-allocated array. Maybe some magic that into_array is doing? Oh well, I'm on board if it works.

shssoichiro (approved):

av1an-core/src/encoder.rs
redzic: A `String` in Rust is just a pointer, length, and capacity, so just 24 bytes (on a 64-bit architecture where the size of `usize` is 64 bits). The stuff that is being stored on the stack is just those 24 bytes for each element, not the actual content of the string (which is still stored on the heap), which is why this works.

Coming to think of it, we could also elide some of the heap allocations of the contents of the strings as well if we use `Cow<str>`.

shssoichiro: Apparently this is another one of those things where I thought at some point in the past I tried to do this and it didn't work.

Godbolt is calling me crazy again.

PR #509, opened 2022-01-04
Remove ctrlc dependency

redzic:
Turns out the behavior is pretty much the same either way. This is probably left over from the initial port from python to rust, and as the rust version does not try to catch some sort of exception or something when pressing ctrl c, only one ctrl c is required for it to stop (a stark contrast to the python version).


master-of-zen (approved):
LGTM

PR #519, opened 2022-01-07
Display possible vspipe errors when ffmpeg pipe is present

redzic:
This makes better error messages in the case of an ffmpeg pipe, where it would otherwise only display `pipe:: invalid data found when processing input` and the error from the encoder.

Here is an error message with this PR, showing the error from all 3 places:

```
ERROR [av1an_core::broker] [chunk 0] encoder crashed: exit status: 1
stdout:

stderr:
        Fatal: Specify stream dimensions with --width (-w)  and --height (-h)


source pipe stderr:
        Deprecated option --y4m specified, use -c y4m instead
        Error: can only apply y4m headers to YUV and Gray format clips without alpha
        Output 150 frames in 68280.21 seconds (0.00 fps)

ffmpeg pipe stderr:
        pipe:: Invalid data found when processing input

```

### DIFF
--- a/av1an-core/src/broker.rs
+++ b/av1an-core/src/broker.rs
@@ -83,7 +83,8 @@ pub struct EncoderCrash {
   pub exit_status: ExitStatus,
   pub stdout: StringOrBytes,
   pub stderr: StringOrBytes,
-  pub pipe_stderr: StringOrBytes,
+  pub source_pipe_stderr: StringOrBytes,
+  pub ffmpeg_pipe_stderr: Option<StringOrBytes>,
 }
 
 impl Display for EncoderCrash {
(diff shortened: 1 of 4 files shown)


shssoichiro (approved):
🏆

shssoichiro:
Oh wait. Clippy is mad.

redzic:
Ah, I was getting an insane amount of borrow checker errors when implementing this, so I switched to manually doing `Option::map` but it may not be necessary anymore. I'll see if I can fix it.

shssoichiro:
Ping on Discord if you need help 👍

redzic:
Should be fixed now, the real fix was to use `Option::take` for certain things. Sometimes clippy does suggest things that the borrow checker does not like though.

shssoichiro:
Particularly since we have pedantic and nursery lints enabled, which tend to be buggier and have more false positives... which is another discussion. 🙂

PR #521, opened 2022-01-08
Fixed mkvmerge on windows by allowing for commands over 32767 characters to be executed.

KesavaViswanadha:


av1an-core/src/concat.rs
```diff
@@ -197,54 +197,10 @@ pub fn mkvmerge(
     .map(|chunk| PathBuf::from(format!("{:05}.{}", chunk, encoder.output_extension())))
     .collect();
 
+  let mut this_file: File = File::create("options.json").unwrap();
```
shssoichiro, line 200: We probably want to create this file in the temp folder.

```suggestion
  let options_path = PathBuf::from(&temp_dir).join("options.json");
  let mut this_file: File = File::create(&options_path).unwrap();
```

av1an-core/src/concat.rs
```diff
@@ -264,6 +220,19 @@ pub fn mkvmerge(
   Ok(())
 }
 
+/** This code adds the necessary command parameters to the options FILE
+ * that is passed in, up to the chunk NUM. Returns nothing.
+ */
+pub fn concatenate_mkvmerge(num: u64, mut this_file: File) {
+  let mut file_string: String = "[ \"-o\", \"output.mp4\", \"audio.mp3\", \"[\"".into();
```
shssoichiro, line 227: We want to make sure we're using the same paths as before, e.g. where we were using `fix_path(output)`.

av1an-core/src/concat.rs
```diff
@@ -264,6 +220,19 @@ pub fn mkvmerge(
...
+  for i in 0..num {
+    file_string.push_str(&format!(" , \"{:05}.ivf\" ", i));
+  }
+  file_string.push_str(",\"]\" ]");
+  println!("{}", file_string);
```
shssoichiro, line 232: I assume this was from debugging, we'll want to make sure we remove it before merging.

av1an-core/src/concat.rs
```diff
@@ -264,6 +220,19 @@ pub fn mkvmerge(
...
+  this_file.write_all(file_string.as_bytes());
```
shssoichiro, line 233: This needs a `.unwrap()` after it in order to ensure it flushes. (Rust's `write` functions are weird like that.)

redzic:
I went ahead and just fixed the errors myself. It now works for me when I try mkvmerge. I guess it's fine to create the options file on all platforms, just in case there are problems with other shells.

redzic:
Thanks for working on this @KesavaViswanadha! I didn't even know mkvmerge supported an options file like ffmpeg until yesterday when I reported the "issue" on their issue tracker (https://gitlab.com/mbunkus/mkvtoolnix/-/issues/3252).

redzic:
@shssoichiro Did you have any other concerns before merging?

shssoichiro (approved):
LGTM

PR #532, opened 2022-01-15
Encoder mode/rate check

master-of-zen:
This pr check does user provide `--end-usage` and rate control values `--cq-level` or `--target-bitrate` when using custom commands and prints warning if they don't

.

### DIFF
--- a/av1an-core/src/settings.rs
+++ b/av1an-core/src/settings.rs
@@ -628,9 +628,63 @@ properly into a mkv file. Specify mkvmerge as the concatenation method by settin
 
     if !self.force {
       self.validate_encoder_params();
+      self.check_rate_control();
     }
 
     Ok(())
+  }
+
+  /// Checks if video parameters have rate control
+  /// and cq-level or target-bitrate set
+  /// Prints the warning
+  fn check_rate_control(&self) {
+    if self.encoder == Encoder::aom {
+      if !self
+        .video_params
+        .iter()
+        .any(|f| Self::check_aom_encoder_mode(f))
+      {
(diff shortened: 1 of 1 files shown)


av1an-core/src/settings.rs
```diff
@@ -628,11 +628,68 @@ properly into a mkv file. Specify mkvmerge as the concatenation method by settin
...
+    }
+
+    s.as_bytes()[END_USAGE.len()..]
+      .iter()
+      .all(|&b| (b as char).is_ascii_alphabetic())
+  }
+
+  fn check_rate(s: &str) -> bool {
+    const CQ_LEVEL: &str = "--cq-level=";
+    const TARGET_BITRATE: &str = "--target_bitrate=";
```
redzic, line 676: This should be with a dash, i.e. `--target-bitrate=`.

av1an-core/src/settings.rs
```diff
@@ -628,11 +628,61 @@ properly into a mkv file. Specify mkvmerge as the concatenation method by settin
...
+    if s.len() <= END_USAGE.len() || !s.starts_with(END_USAGE) {
+      return false;
+    }
+
+    s.as_bytes()[END_USAGE.len()..]
+      .iter()
+      .all(|&b| (b as char).is_ascii_alphabetic())
+  }
+
+  fn check_rate(s: &str) -> bool {
```
redzic, line 667: It feels like this function and `check_end_usage` should be renamed to reflect that it only applies to aomenc.

redzic (approved):

PR #546, opened 2022-01-17
Rework log levels to be more useful

shssoichiro:
- Update log levels so that the log file contains more verbose
  information by default, while leaving the stderr progress indicator
  relatively uncluttered
- Explicitly set the log level to `error` for all dependencies
- rav1e scenechange info has been moved to `--log-level trace`
- Avoid outputting colors if stderr is not a tty
- Fix a crash when using `--quiet`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -156,6 +156,7 @@ version = "0.3.1"
 dependencies = [
  "ansi_term",
  "anyhow",
+ "atty",
  "av1an-core",
  "clap",
  "ffmpeg-next",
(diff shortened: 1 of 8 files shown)


redzic:
It seems that there is still some colored info printed while not running in a tty, namely the params, warn for rate control for aomenc, just the message for "Scene detection", and the queue, workers, passes message. Perhaps we should fix that too?

shssoichiro:
@redzic Ping. I pushed up the changes you requested.

redzic (approved):
Looks good, thanks!

PR #550, opened 2022-01-19
Display chunk index instead of worker index in verbose mode

shssoichiro:
The chunk index would be much more useful for users who are using
verbose mode. Worker index is easily derived from counting the lines,
and is not particularly useful.

Screenshot 2022-01-18 194916

### DIFF
--- a/av1an-core/src/progress_bar.rs
+++ b/av1an-core/src/progress_bar.rs
@@ -180,10 +180,9 @@ pub fn init_multi_progress_bar(len: u64, workers: usize, total_chunks: usize) {
   });
 }
 
-pub fn update_mp_chunk(worker_idx: usize, chunk: usize, total_chunks: usize) {
-  let digits = printable_base10_digits(total_chunks) as usize;
+pub fn update_mp_chunk(worker_idx: usize, chunk: usize, padding: usize) {
   if let Some((_, pbs)) = MULTI_PROGRESS_BAR.get() {
-    pbs[worker_idx].set_prefix(format!("[Chunk {:>digits$}]", chunk, digits = digits));
+    pbs[worker_idx].set_prefix(format!("[Chunk {:>digits$}]", chunk, digits = padding));
   }
 }
 
--- a/av1an-core/src/progress_bar.rs
+++ b/av1an-core/src/progress_bar.rs
@@ -144,15 +144,15 @@ pub fn reset_mp_bar_at(pos: u64) {
   }
 }
 
-pub fn init_multi_progress_bar(len: u64, workers: usize) {
+pub fn init_multi_progress_bar(len: u64, workers: usize, total_chunks: usize) {
   MULTI_PROGRESS_BAR.get_or_init(|| {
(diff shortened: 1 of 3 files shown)


av1an-core/src/progress_bar.rs
```diff
@@ -144,23 +144,23 @@ pub fn reset_mp_bar_at(pos: u64) {
...
+    for _ in 1..=workers {
       let pb = ProgressBar::hidden()
         // no spinner on windows, so we remove the prefix to line up with the progress bar
         .with_style(ProgressStyle::default_spinner().template(if cfg!(windows) {
           "{prefix:.dim} {msg}"
         } else {
           "  {prefix:.dim} {msg}"
         }));
-      pb.set_prefix(format!("[Worker {:>digits$}]", i, digits = digits));
+      pb.set_prefix(format!("[Idle  {}]", " ".repeat(digits)));
```
redzic, line 163: I believe this can be done with a single `format!` like so, avoiding extra allocations: `format!("{:width$}]", "[Idle ", width = digits)`

redzic: Oops, I commented on the wrong thing.

shssoichiro: Wait, did you? I think this is a valid improvement

shssoichiro: It turns out, I'm not sure if we ever actually _show_ this message. But I'll fix it anyway just in case 😅

redzic: I remember seeing a message similar to `[Idle]` in rav1e-by-gop, but I guess av1an has no time to waste on idle workers 💪

redzic (approved):
Looks like a great improvement, thanks!

av1an-core/src/progress_bar.rs
```diff
@@ -144,23 +144,23 @@ pub fn reset_mp_bar_at(pos: u64) {
...
 
-pub fn init_multi_progress_bar(len: u64, workers: usize) {
+pub fn init_multi_progress_bar(len: u64, workers: usize, total_chunks: usize) {
   MULTI_PROGRESS_BAR.get_or_init(|| {
     let mpb = MultiProgress::new();
 
     let mut pbs = Vec::new();
 
-    let digits = printable_base10_digits(workers) as usize;
+    let digits = printable_base10_digits(total_chunks) as usize;
```
redzic, line 153: Oh one very small thing actually. This should be `total_chunks.saturating_sub(1)` (in case of overflow I guess?) since we display the index.

shssoichiro: You are technically correct. The best kind of correct. (the outcome of not fixing it is if there are exactly 10 or 100 or 1000 chunks, there's an extra space. I'll still open another PR to fix it)

redzic: Ah, you don't need to worry about it. I already pushed a commit to this PR that fixes it :)

PR #562, opened 2022-01-31
Future proof SVT-AV1 version parsing

redzic:
The old code will not work for future versions that use the same syntax, for example if there is a 0.9.1 version, as we only check for 0.9.0.


redzic:
I just realized this won't work for versions that don't correctly into base 10 numbers (e.g. 0.8.11, even though there is no such version of svt). Marking this as WIP.

shssoichiro (approved):

PR #568, opened 2022-02-06
Update several dependencies

shssoichiro:
Notably:
- parking_lot
- sysinfo
and many patch version bumps

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,7 +187,6 @@ dependencies = [
  "av-ivf",
  "av-scenechange",
  "cfg-if 1.0.0",
- "chrono",
  "crossbeam-channel",
  "crossbeam-utils",
  "dashmap",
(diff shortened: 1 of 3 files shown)


redzic:
It seems like updating indicatif is somehow causing the tests to fail?

redzic:
That particular panic seems like it would be fixed by https://github.com/console-rs/indicatif/commit/5afa9a4a789d6a74586c0e2d27ae5bbe2cb8c460.

@shssoichiro Should we just avoid updating indicatif until they make a new release containing the fix, or should we go ahead and use the git version for indicatif until it makes it to a release?

shssoichiro:
Interesting. Since we're already using git for other things, I think it's fine to use the git version for indicatif for now as well.

redzic (approved):

Issue #573, opened 2022-02-16
[feature request] Scene detection only run.


lastrosade:
This would allow me to avoid running scene detection very slowly with a vpy script or to bodge it with incorrect setting to force av1an to crash.

This way I could run scene detection fast with the source file before encoding with a vpy script.

Something along the lines of `av1an -i video.mkv -s video.json --sc-only --sc-method=fast`

lavalamp3774:
It would also be nice if this were multi-threaded. Currently I'm running scene detection on a 4K movie at only 23 fps on a single core. Without --sc-method=fast I get about 12 fps.

redzic:
@lavalamp3774 There is also the `--sc-downscale-height` option which can speed up scene detection. Something like `--sc-pix-format yuv420p --sc-method fast --sc-downsale-height 720` should be faster for you.

shssoichiro:
Although it's important/unfortunate to note that at 4k and higher resolutions, scene detection (especially fast mode) becomes increasingly bottlenecked by decoding. The tips above will help, but only by so much, unfortunately.

silverbacknet:
@redzic This is more for multi resolutions or qualities, particularly for test encodes, I think, or perhaps for pipelining long queues so that there aren't long periods of most cores being idle between encodes while scene detection is running. It can be shoved to a core or two to handle ahead of time while the bulk are encoding the prior video.

PR #574, opened 2022-02-16
Lastrosade sc only

lastrosade:
This pull request adds the `--sc-only` boolean argument which forces av1an to stop after scene detection as per #573 

Right now it simply deletes the temporary directory after the run

What would be optimal would be to not create any temp dir if both:
1. The input is a vpy script.
2. -s is specified.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -225,7 +225,6 @@ jobs:
       - name: Testing sc-only
         run: |
           target/ci/av1an -i tt_sif.y4m --sc-only --sc-method fast -s "tt_sif.json"
-          rm tt_sif.json
           du -h tt_sif.json
 
   docker:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -225,7 +225,7 @@ jobs:
       - name: Testing sc-only
         run: |
           target/ci/av1an -i tt_sif.y4m --sc-only --sc-method fast -s "tt_sif.json"
-          du -h tt_sif.json
+        if: du -h tt_sif.json || cat tt_sif.json
 
   docker:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


redzic:
It looks like the formatting check failed, run `cargo fmt` to fix the formatting and then push the commit

lastrosade:
Do I need to write unit tests?

redzic:
I'm not exactly sure how you would write a unit test, but we can add an integration test instead (which are in `.github/workflows/tests.yml`).

redzic:
Perhaps we should require `-s` to be specified if `--sc-only` is set?

lastrosade:
Yes, let me try to figure that out.

redzic:
@lastrosade The easiest way to do it would probably be with this method: https://docs.rs/clap/3.1.0/clap/struct.Arg.html#method.requires_if

lastrosade:
This is a train wreck

lastrosade:
It seems that requires_if only wants strings

lastrosade:
I figured it out, it was dumb.

lastrosade:
The integration test failed with `Error: Failed to parse scenes file "tt_sif.json", this likely means that the scenes file is corrupted` I doubt this has anything to do with my patch.

Is the video generating an empty scene file?

lastrosade:
running the same test on the same file works fine for me.

shssoichiro:
It seems the error is coming from this branch:

```rust
let (mut scenes, frames) = if (self.scenes.is_some() && scene_file.exists()) || self.resume {
  used_existing_cuts = true;
  crate::split::read_scenes_from_file(scene_file.as_ref())?
}
```

Which should only get hit if the scenes file already exists prior to running your test command. It shouldn't be... so this is a bit weird.

Just to prove we're not insane, could you try to `rm tt_sif.json` prior to calling av1an in your CI test?

lastrosade:
It did not fix it

master-of-zen (approved):
Looks good to me

PR #575, opened 2022-02-18
Update to ffmpeg 5.0

redzic:
- Also rename ffmpeg-next in Cargo.toml, and refer to our own ffmpeg module by `crate::ffmpeg`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -501,8 +501,7 @@ dependencies = [
 [[package]]
 name = "ffmpeg-next"
 version = "5.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "70a1fc87054b079ade0b081b023c5fef309cd4bbcb569c09eed2aa39b25a0a11"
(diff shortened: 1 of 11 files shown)


shssoichiro:
Tests do not seem to like that...

> Error: ffmpeg: failed to get pixel format for input video "tt_sif.y4m"

shssoichiro:
Does the CI runner also need to be on ffmpeg 5? Could that potentially cause this issue?

redzic:
Actually I can reproduce the CI failure locally as well... so I'm not sure exactly what's wrong. I'll try to debug the issue further.

redzic:
It now runs fine for me locally (and `all-tests` seem to pass as well), but it looks like there's an issue with compiling on the docker container?

master-of-zen:
Works now))

master-of-zen (approved):
Great job :)

svenstaro:
@master-of-zen Could we get a new release containing this update? I'd like to build av1an against ffmpeg in Arch in order to prepare for the drop of ffmpeg4.4 and av1an is currently one of the blocking packages.

shssoichiro:
It looks like we'll need new releases of rav1e, av-scenechange, vapoursynth-rs, and console, since we are still depending on git versions of those. I can push for a release on the first two. Not sure about the latter two.

Issue #220, opened 2021-01-26
[feature request] Multiple input at the same time


lastrosade:
Say we have a directory with 1000+ small videos that could be split in 3 to 4 parts

It would be neat if av1an could just scan them all one after the other and slowly add each parts to the worker queue.

This would allow a major speedup when encoding a lot of small content like camera rolls and such.

lastrosade:
In order to select settings per video, we could provide scripts to av1an. For example
For example "-ff script.ps1" with:
```powershell
#!/bin/pwsh
param($path)

if ([int](mediainfo --Output="Video;%Height%" "$path") -ge [int](mediainfo --Output="Video;%Width%" "$path")) {
    echo = "-vf 'scale=min(1080\,ih):-1:flags=sinc'"
} else {
    echo = "-vf 'scale=-1:min(1080\,iw):flags=sinc'"
}
```
Here, av1an would run the script on every videos in the directory to get the desired settings.

Could this work around the issues we discussed before?

lastrosade:
I missed this issue's second birthday :(

PR #577, opened 2022-02-21
Multiple input files

master-of-zen:
Allows to use `-i` to give av1an multiple files.
Allows to use `-i` as input for folders.


lastrosade:
Would this address some parts of #220 ?

av1an-cli/src/lib.rs
```diff
@@ -442,164 +442,212 @@ fn confirm(prompt: &str) -> io::Result<bool> {
...
-pub fn parse_cli(args: CliOpts) -> anyhow::Result<EncodeArgs> {
-  let temp = if let Some(path) = args.temp.as_ref() {
-    path.to_str().unwrap().to_owned()
-  } else {
-    format!(".{}", hash_path(args.input.as_path()))
-  };
+/// Given Folder and File paths as inputs
+/// Converts them all to file paths
+/// Converting only depth 1 of Folder paths
+pub fn convert_input(args_paths: Vec<PathBuf>) -> Vec<PathBuf> {
```
redzic, line 448: I don't think it's necessary to take a `Vec<PathBuf>` by here, since we're not actually consuming the value (only iterating over it)

av1an-cli/src/lib.rs
```diff
@@ -442,164 +442,212 @@ fn confirm(prompt: &str) -> io::Result<bool> {
...
+              Some(d.path())
+            } else {
+              None
+            }
+          } else {
+            None
+          }
+        })
+        .collect();
+      validated_inputs.extend(files)
```
redzic, line 471: Since this is an iterator, I think you can extend without an intermediate vector/allocation.

av1an-cli/src/lib.rs
```diff
@@ -442,164 +442,212 @@ fn confirm(prompt: &str) -> io::Result<bool> {
...
+/// Converting only depth 1 of Folder paths
+pub fn convert_input(args_paths: Vec<PathBuf>) -> Vec<PathBuf> {
+  let mut validated_inputs: Vec<PathBuf> = Vec::with_capacity(args_paths.len());
+
+  for input_arg in args_paths {
+    if input_arg.is_file() {
+      validated_inputs.push(input_arg.to_path_buf())
+    } else if input_arg.is_dir() {
+      let files: Vec<PathBuf> = std::fs::read_dir(input_arg)
+        .unwrap()
```
redzic, line 456: This should probably return an error rather than unwrapping.

av1an-core/src/lib.rs
```diff
@@ -241,7 +241,7 @@ pub fn list_index(params: &[impl AsRef<str>], is_match: fn(&str) -> bool) -> Opt
   })
 }
 
-#[derive(Serialize, Deserialize, Debug, EnumString, IntoStaticStr, Display)]
+#[derive(Serialize, Deserialize, Debug, EnumString, IntoStaticStr, Display, Clone)]
```
redzic, line 244: This should probably also implement `Copy`, because it's a trivial enum.

av1an-cli/src/lib.rs
```diff
@@ -657,10 +705,12 @@ impl LogWriter for StderrLogger {
   }
 }
 
+pub fn validate_input_files() {}
```
redzic, line 708: Why is there an empty function here?

av1an-cli/src/lib.rs
```diff
@@ -442,164 +442,212 @@ fn confirm(prompt: &str) -> io::Result<bool> {
...
-      args.encoder.get_default_pass()
-    },
-    video_params: if let Some(args) = args.video_params.as_ref() {
-      shlex::split(args).ok_or_else(|| anyhow!("Failed to split video encoder arguments"))?
+/// Returns vector of Encode args ready to be fed to encoder
+pub fn parse_cli(args: CliOpts) -> anyhow::Result<Vec<EncodeArgs>> {
+  let test = convert_input(args.input.clone());
+
+  if test.is_empty() {
+    anyhow!("No valid input file");
```
redzic, line 483: I'm not sure if this actually bubbles up the error, since `anyhow!` just creates the error struct, but does not return it immediately. I think you just need a `return Err(anyhow!( ... ))`.

redzic:
I think we should support some way of specifying custom output locations for each input file, rather than relying on it all being the default output file name.

av1an-cli/src/lib.rs
shssoichiro: The anyhow crate also has a `bail` macro that does this in one step: `bail!("No valid input file");`

master-of-zen:
rebasing is amazingm i have everything work on my side

av1an-cli/src/lib.rs
```diff
@@ -442,164 +442,217 @@ fn confirm(prompt: &str) -> io::Result<bool> {
...
+          } else {
+            None
+          }
+        })
+        .collect()
+    })
   } else {
-    format!(".{}", hash_path(args.input.as_path()))
-  };
+    Ok(vec![])
```
redzic, line 474: This should probably return an error instead of an empty vector, instead of relying on the caller to check for the empty case.

av1an-cli/src/lib.rs
```diff
@@ -442,164 +442,217 @@ fn confirm(prompt: &str) -> io::Result<bool> {
...
+pub fn convert_input(path: &Path) -> anyhow::Result<Vec<PathBuf>> {
+  // TODO: to validate file extensions
+  // let valid_media_extensions = ["mkv", "mov", "mp4", "webm", "avi", "qt", "ts", "m2t", "py", "vpy"];
+
+  if path.is_file() {
+    Ok(vec![path.to_path_buf()])
+  } else if path.is_dir() {
+    Ok({
+      let dir = std::fs::read_dir(path)?;
+      dir
```
redzic, line 457: Since this also exists in other places in the code, I think this should be its own function.

master-of-zen:
> Would this address some parts of #220 ?
Yep, it's a step in that direction)

PR #580, opened 2022-02-26
Update dependencies

redzic:


redzic:
Somehow this is causing an extremely bizarre issue with pausing during scene detection. Closing until I can fix it.

PR #582, opened 2022-03-02
Multiple input files

master-of-zen:


av1an-cli/src/lib.rs
```diff
@@ -691,19 +725,22 @@ pub fn run() -> anyhow::Result<()> {
   // and then it calls their shutdown method.
   let _logger = Logger::with(log)
     .log_to_file_and_writer(
-      FileSpec::try_from(PathAbs::new(&args.log_file)?)?,
+      // UGLY: take first or the files for log path
```
shssoichiro, line 728: Maybe we should move the log file initialization to within the input files loop? Then each input file can write to its own log file.

master-of-zen: Yep, current code is quite intertwined, a lot of stuff require to be more independently initialized

PR #584, opened 2022-03-05
Adaptive extra splits

master-of-zen:
Sets extra splits to 10 * fps if extra splits option was not provided, defaults to 240 if failed to get fps

### DIFF
--- a/av1an-cli/src/lib.rs
+++ b/av1an-cli/src/lib.rs
@@ -510,13 +510,8 @@ pub fn parse_cli(args: CliOpts) -> anyhow::Result<EncodeArgs> {
     concat: args.concat,
     encoder: args.encoder,
     extra_splits_len: match args.extra_split {
-      Some(x) => {
-        if x > 0 {
-          Some(x)
-        } else {
-          None
-        }
-      }
+      Some(0) => None,
+      Some(x) => Some(x),
       // Make sure it's at least 10 seconds
       None => match input.frame_rate() {
         Ok(fps) => Some((fps * 10.0) as usize),
--- a/av1an-cli/src/lib.rs
+++ b/av1an-cli/src/lib.rs
@@ -609,11 +609,11 @@ pub fn parse_cli(args: CliOpts) -> anyhow::Result<EncodeArgs> {
   Ok(encode_args)
 }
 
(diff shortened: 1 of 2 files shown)


av1an-cli/src/lib.rs
```diff
@@ -609,6 +609,14 @@ pub fn parse_cli(args: CliOpts) -> anyhow::Result<EncodeArgs> {
   Ok(encode_args)
 }
 
+///  Make make sure splits are not longer than 10 seconds
+pub fn calculate_extra_splits_from_fps(fps: anyhow::Result<f64>) -> usize {
```
redzic, line 613: This function probably doesn't need to exist, can we move this code into where the function is being called?

av1an-cli/src/lib.rs
```diff
@@ -514,10 +509,19 @@ pub fn parse_cli(args: CliOpts) -> anyhow::Result<EncodeArgs> {
     chunk_order: args.chunk_order,
     concat: args.concat,
     encoder: args.encoder,
-    extra_splits_len: if args.extra_split > 0 {
-      Some(args.extra_split)
-    } else {
-      None
+    extra_splits_len: match args.extra_split {
+      Some(x) => {
```
redzic, line 513: This can be written with less lines like this:

```rust
match args.extra_split {
  Some(0) => None,
  Some(x) => Some(x),
  None => { /* rest of the code */ }
}
```

av1an-cli/src/lib.rs
```diff
@@ -514,10 +509,19 @@ pub fn parse_cli(args: CliOpts) -> anyhow::Result<EncodeArgs> {
...
+        if x > 0 {
+          Some(x)
+        } else {
+          None
+        }
+      }
+      // Make sure it's at least 10 seconds
+      None => match input.frame_rate() {
+        Ok(fps) => Some((fps * 10.0) as usize),
```
redzic, line 522: We probably want to cap the maximum to some reasonable value so as to not choose unreasonably high values for high framerate content.

master-of-zen: 10 seconds seems reasonable to tome

redzic (approved):

PR #591, opened 2022-03-14
Speed up CI, Reduce container image size

luigi311:
Use caching for cargo build to reduce the build times from 5 mins to 45 seconds when dependencies are non changing

Restructure the entire Dockerfile to use multistaging to reduce the docker image by ~500 mb from 1.05 GB to 584 MB.
Use github action internal caching so we dont have to deal with uploading and download the cache as an artifact.
Use cargo chef to build all the dependencies as a layer so that can be cached and have to only compile av1an itself.

Without cargo chef after my initial changes cold run was 9m 49s and when caching the build times came down to 6m 28s but the cache would be invalided on any code change so it was not usable
With cargo chef on a cold run it was still 9m 49s and a cached run it came down to 4m 57s, since this  is caching the dependencies and still building av1an the time going forward should be closer to the 5m mark than the 9m 49s cold run.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,31 @@
+FROM archlinux:base-devel AS planner
+RUN pacman -Syy --noconfirm
+
+# Install all dependencies (except for rav1e)
+RUN pacman -S --noconfirm rsync rust clang nasm git aom ffmpeg vapoursynth ffms2 libvpx mkvtoolnix-cli svt-av1 vapoursynth-plugin-lsmashsource vmaf
+
+WORKDIR /tmp/Av1an
+RUN cargo install cargo-chef 
+COPY . .
+RUN cargo chef prepare  --recipe-path recipe.json
+
+
+
+
+FROM archlinux:base-devel AS cacher
+RUN pacman -Syy --noconfirm
+
+# Install all dependencies (except for rav1e)
+RUN pacman -S --noconfirm rsync rust clang nasm git aom ffmpeg vapoursynth ffms2 libvpx mkvtoolnix-cli svt-av1 vapoursynth-plugin-lsmashsource vmaf
+
+WORKDIR /tmp/Av1an
+RUN cargo install cargo-chef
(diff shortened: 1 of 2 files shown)


master-of-zen:
Amazing 🤩

Dockerfile
```diff
@@ -1,36 +1,71 @@
...
-
-# Copy av1an and build av1an
-COPY --chown=app_user . /Av1an
-WORKDIR /Av1an
-RUN cargo build --release && \
-    mv ./target/release/av1an /usr/local/bin && \
-    cd .. && rm -rf ./Av1an
-
-# Remove build dependencies
-RUN pacman -R --noconfirm rust clang nasm git
```
redzic, line 33: Is it possible to add back this line? The build dependencies are not needed at runtime, so the container size could be further reduced.

Dockerfile
```diff
@@ -1,36 +1,71 @@
-FROM archlinux:base-devel
+FROM archlinux:base-devel AS planner
+RUN pacman -Syy --noconfirm
 
-ENV MPLCONFIGDIR="/home/app_user/"
-ARG DEPENDENCIES="mkvtoolnix curl llvm clang"
+# Install all dependencies (except for rav1e)
+RUN pacman -S --noconfirm rsync rust clang nasm git aom ffmpeg vapoursynth ffms2 libvpx mkvtoolnix-cli svt-av1 vapoursynth-plugin-lsmashsource vmaf
```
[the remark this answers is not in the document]
luigi311: This uses multistaging which are separate instances of the container. Since they are separate you just need to install what you need for each stage.

Dockerfile
luigi311: Removing anything afterwards doesnt affect the size of the docker container as those still exist in the layers prior due to how docker works. Those build dependencies actually do not exist in the final runtime due to them only existing in the build stage and is the benefit of using multistaging.

redzic (approved):

PR #595, opened 2022-03-17
Silence remaining clippy lints

shssoichiro:
We intentionally transmute here for performance reasons.


redzic (approved):

redzic:
Thanks, I'm not even sure what the alternative here since this is literally how the documentation tells you to use uninitialized memory.

shssoichiro:
Supposedly this lint has a lot of issues and false positives, which is why it was moved to the nursery.

PR #596, opened 2022-03-18
Fix stalled pipe on Windows when using target quality

woot000:
This is my first time making changes to Rust code, so let me know if the styling is wrong or if I made a glaring mistake with this change. It works perfectly on my machine

### DIFF
--- a/av1an-core/src/target_quality.rs
+++ b/av1an-core/src/target_quality.rs
@@ -187,7 +187,7 @@ impl<'a> TargetQuality<'a> {
       let mut source = if let [pipe_cmd, args @ ..] = &*chunk.source {
         tokio::process::Command::new(pipe_cmd)
           .args(args)
-          .stderr(Stdio::piped())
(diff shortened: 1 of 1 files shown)


redzic:
The cause of this problem appears to be related to https://github.com/rust-lang/rust/issues/45572, where the pipe buffer becomes full, blocking further output. The only problem I have with this approach is that we lose the error reporting from the stderr of the child processes, so debugging any issues with target quality crashes would become a big pain without any error messages to go off of.

Perhaps we could look into some other solution for handling this? I would ideally want something that fixes this while keeping the error reporting that we currently have.

woot000:
I suppose stderr could be redirected to a file, though that would necessitate a lot of changes to the vmaf_probe function, which is out of the scope of my abilities for Rust coding

I think for the time being, this is okay as a temporary fix

redzic (approved):
Sorry for the extremely slow response, I'll merge this as a temporary fix while I look for a solution that has proper error handling/reporting.

PR #598, opened 2022-03-21
Add zones support

### LINKED ISSUE #267: [Feature Request] Change encoding parameters for specific segments
I'm still thinking through what the implementation for this might look like--maybe "Provide a list of frame ranges, and if the start frame for a segment is within that range, use these parameters instead of the default ones for the encode"

My use case here is: I may have segments of some files where I want to change encoding settings for a specific segment. For example, if there is a segment of a video with very heavy film grain, I may want to adjust the cq level or other settings to encode that segment differently from the rest of the file.

shssoichiro:
Currently implemented:
- Overriding the selected encoder & number of passes
- Overriding or adding video params
- Overriding photon noise setting
- Overriding min/max scene length

Closes #267

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,7 +135,7 @@ dependencies = [
 [[package]]
 name = "av-scenechange"
 version = "0.7.2"
-source = "git+https://github.com/rust-av/av-scenechange?rev=9425e3597b27c79cd970756b7cdb9c24cb2c4d03#9425e3597b27c79cd970756b7cdb9c24cb2c4d03"
+source = "git+https://github.com/rust-av/av-scenechange?rev=0c63f493200eeccc21193d129fc63cb1265e40fa#0c63f493200eeccc21193d129fc63cb1265e40fa"
 dependencies = [
  "anyhow",
  "rav1e",
@@ -195,6 +195,7 @@ dependencies = [
  "itertools",
  "log",
  "memchr",
+ "nom",
  "num_cpus",
  "once_cell",
  "parking_lot",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ debug-assertions = true
 overflow-checks = true
 
 [patch.crates-io]
-av-scenechange = { git = "https://github.com/rust-av/av-scenechange", rev = "9425e3597b27c79cd970756b7cdb9c24cb2c4d03" }
+av-scenechange = { git = "https://github.com/rust-av/av-scenechange", rev = "0c63f493200eeccc21193d129fc63cb1265e40fa" }
 # TODO: switch to release version once the fix for av_get_best_stream is published on crates.io.
 ffmpeg-next = { git = "https://github.com/zmwangx/rust-ffmpeg", rev = "0054b0e51b35ed240b193c7a93455714b4d75726" }
 console = { git = "https://github.com/console-rs/console", rev = "5484ea9d9f6884f6685349708e27bf08fab7703c" }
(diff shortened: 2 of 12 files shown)


master-of-zen:
Looks amazing

redzic:
It looks like there is no protection against the user specifying a different encoder which conflicts with the main encoder, for example specifying x264 with aomenc. Is it possible to add error handling for that situation? Somehow in that case the concatenation even ends up succeeding, but the output file ends up being silently broken. Also, since we're parsing untrusted input, it would be nice to look into some sort of fuzzing solution.

redzic:
For some reason I get a panic with these parameters on a 24 frame video:

```
redzic@threadripper ~/P/Av1an (zones)> ./target/release/av1an -i ~/input.y4m --zones zones.txt -e rav1e
Scene detection
INFO [av1an_core::settings] scenecut: found 3 scene(s) [with extra_splits (239 frames): 3 scene(s)]
  00:00:00 [##############################################################################################################################################] 100%  24/24 (37.36 fps, eta 0s)Queue 2 Workers 2 Passes 1
Params: --speed 6 --quantizer 100 --no-scene-detection
⠦ 00:00:01 [---------------------------------------------------------------------------------------------------------------------------------------------------]   0%  0/24 (0 fps, eta 0s)thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ffmpeg::Error(2: No such file or directory)', av1an-core/src/broker.rs:277:70
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Any { .. }', av1an-core/src/broker.rs:188:35
⠋ 00:00:05 [##############>-------------------------------------------------------------------------------------------------]  12%  3/24 (3.13 s/fr, eta 66s, 3532.9 Kbps, est. 431.69 KiB)thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Any { .. }', av1an-core/src/settings.rs:1241:21
```

Where this is the contents of `zones.txt`:
```
3 24 aom reset --cq-level=30
```

Can you see if you can reproduce this issue as well?

Also, I encountered another edge case where there is no check for if all encoders support the pixel format specified by `--pix-format`, which can lead to unexpected errors. For example, this leads to an error when trying to implicitly encode yuv444p with SVT-AV1:

```
redzic@threadripper ~/P/Av1an (zones)> ./target/release/av1an -i ~/input.y4m --pix-format yuv444p --zones zones.txt 
Default output file "input_aom.mkv" exists. Do you want to overwrite it? [Y/n]: y
Scene detection
INFO [av1an_core::settings] scenecut: found 3 scene(s) [with extra_splits (239 frames): 3 scene(s)]
  00:00:00 [##############################################################################################################################################] 100%  24/24 (40.99 fps, eta 0s)Queue 2 Workers 2 Passes 2
Params: --threads=8 --cpu-used=6 --end-usage=q --cq-level=30
WARN [av1an_core::broker] Encoder failed (on chunk 1):
encoder crashed: exit status: 1
stdout:

stderr:
        Svt[info]: -------------------------------------------
        Svt[info]: SVT [version]:       SVT-AV1 Encoder Lib v0.9.0-dirty
        Svt[info]: SVT [build]  :       GCC 11.2.0       64 bit
        Svt[info]: LIB Build date: Mar  6 2022 15:48:38
        Svt[info]: -------------------------------------------
        Error instance 1: 444 chroma format not supported, only 420 is supported at this time.
        Error found when reading the y4m file parameters.
        Error in configuration, could not begin encoding! ... 
        Run SvtAv1EncApp --help for a list of options
        SvtMalloc[info]: you have no memory leak


source pipe stderr:
        Deprecated option --y4m specified, use -c y4m instead
```

With this `zones.txt`:

```
3 24 svt-av1 reset --crf 21
```

redzic:
Could we also automatically set 1-pass mode for aom/vpx if `--rt` was specified, like how we do with the main encoding?

shssoichiro:
Bump.

master-of-zen:
LGTM

master-of-zen (approved):
Amazing work) Thanks)

PR #606, opened 2022-04-03
Include whether lsmash and ffms2 were found in version info

redzic:
This makes it very easy for the user to tell whether or not lsmash or ffms2 is installed, which can help in debugging/troubleshooting situations as well.


shssoichiro (approved):
nice

PR #612, opened 2022-04-15
Error and retry chunk if frame count mismatches

shssoichiro:
I can't think of a situation where someone would want this issue to be
ignored, so we should treat it as any other encoder error and perform 3
retries, then exit if it continues to fail. This probably indicates some
kind of issue with decoding or applying filters to that chunk.

### DIFF
--- a/av1an-core/src/broker.rs
+++ b/av1an-core/src/broker.rs
@@ -2,7 +2,7 @@ use crate::progress_bar::update_progress_bar_estimates;
 use crate::util::printable_base10_digits;
 use crate::DoneChunk;
 use crate::{
-  ffmpeg, finish_multi_progress_bar, finish_progress_bar, get_done,
+  finish_multi_progress_bar, finish_progress_bar, get_done,
   progress_bar::{dec_bar, dec_mp_bar},
   settings::EncodeArgs,
   Chunk, Encoder, Instant, TargetQuality, Verbosity,
@@ -247,7 +247,7 @@ impl<'a> Broker<'a> {
 
           if r#try == self.max_tries {
             error!(
-              "[chunk {}] encoder crashed {} times, shutting down worker",
+              "[chunk {}] encoder failed {} times, shutting down worker",
               chunk.index, self.max_tries
             );
             return Err(e);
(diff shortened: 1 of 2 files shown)


redzic:
I think we also have to explicitly decrement the progress bar in the case of that type of failure.

av1an-core/src/broker.rs
```diff
@@ -269,52 +269,60 @@ impl<'a> Broker<'a> {
...
+                audio_size_bytes.load(atomic::Ordering::SeqCst),
+              );
+
+              debug!(
+                "finished chunk {:05}: {} frames, {:.2} fps, took {:.2?}",
+                chunk.index, chunk.frames, fps, enc_time
+              );
+            } else {
+              error!(
+                "finished chunk: FRAME MISMATCH: chunk {}: {}/{} (actual/expected frames)",
```
redzic, line 309: Perhaps this should be a warning message instead of an error since we are still restarting the chunk? I think this might be confusing to some users, what do you think though?

shssoichiro: That's fair, it looks like in the other places we warn unless it's at max retries and bails.

shssoichiro:
I happened to have a file that got through the whole encode with one chunk with a significant frame mismatch. It muxed and happily deleted all of my temp files, and the resulting output had a significant desync after the broken chunk. So, just confirming that retrying after this and erroring after 3 failures _is_ the correct approach.

av1an-core/src/broker.rs
```diff
@@ -269,52 +269,60 @@ impl<'a> Broker<'a> {
...
+                audio_size_bytes.load(atomic::Ordering::SeqCst),
+              );
+
+              debug!(
+                "finished chunk {:05}: {} frames, {:.2} fps, took {:.2?}",
+                chunk.index, chunk.frames, fps, enc_time
+              );
+            } else {
+              warn!(
+                "finished chunk: FRAME MISMATCH: chunk {}: {}/{} (actual/expected frames)",
```
redzic, line 309: I think we also have to explicitly return an error from this entire function if we are on the last try, similar to what we do on line 253, since otherwise I think this function will actually not return any type of error at all if the frame count mismatch happens `max_tries` times.

shssoichiro: In order to avoid a lot of this duplication, I moved the error handling logic up into `create_pipes` so we'll just get an Err back from that function and it'll be handled by the existing code here.

redzic:
Looks much better without the code duplication, thanks! My only suggestion is to change line 250 of broker.rs to say "encoder failed" instead of "encoder crashed", to both match our warning message and to reflect our now more generic usage of `EncoderCrash`.

redzic (approved):

PR #614, opened 2022-04-16
docs: First pass for SVT-AV1 documentation

### LINKED ISSUE #592: Please Update  SVT-AV1 docs
SVT-AV1 docs is Empty. Please Update is.
Svt-av1 is the only stable encoder available to me at present.

hansejo:
Adding some essential SVT-AV1 documentation. Requesting feedback before merge.

closes #592

### DIFF
--- a/docs/Encoders/svt-av1.md
+++ b/docs/Encoders/svt-av1.md
@@ -1,1 +1,120 @@
- 
+# SVT-AV1
+
+This will be a quick guide for setting options svt-av1 when using with Av1an. For more complete documentation, you should read the official documentation:
+
+- [SVT-AV1 Docs - GitLab](https://gitlab.com/AOMediaCodec/SVT-AV1/-/tree/master/Docs)
+
+Make sure your svt-av1 encoder is up-to-date.
+
+# Rate control
+
+## --rc
+
+The `--rc` option selects what rate control strategy you want to use.
+
+- `--rc 0` - Constant rate factor
+- `--rc 1` - Variable bit rate
+- `--rc 2` - Constant bit rate
+
+## Constant rate factor 
+
+```
+--rc 0 --crf
+```
+
+Constant rate factor, a.k.a. constant quality. This is the most common way of determining video quality. You set --crf to be a number between 1-63, and the encoder will work out a bit-rate to keep constant quality. The lower the number, the less compression, the higher it is, the more compression will be used. Anything lower than 20 is considered "hi-fi", 30+ is for mini encodes.
+
+**Example:**
+
+```
+av1an ... -v " --rc 0 --crf 24 --preset 4 --input-depth 10 --tune 0" ...
+```
+
+## Variable bit-rate
+
+```
+--rc 1 --tbr
+```
+
+Variable bit-rate. Requires you to set target bit-rate `--tbr`.
+
+**Example:**
+
+```
+av1an ... --passes 2 -v " --rc 1 --tbr 2000 --preset 4 --input-depth 10 --tune 0" ...
+```
+
+# Preset
+
+```
+--preset
+```
+
+If RC controls the compression strategy, then the preset determines what optimisation features get enabled (at the cost of encode time). Realistically the range for preset is 0-13, with 0 being the most optimized (and slow). 0-3 is only for the latest and fastest system, or the most patient of people. 4-6 is more common among enthusiasts. You should go as low as you can bare, 4 is a good starting place for most.
+
+**Example:**
+
+```
+... --preset 4 ...
+```
+
+# Tune
+
+```
+--tune
+```
+
+- `--tune 0` - VQ
+- `--tune 1` - PSNR
+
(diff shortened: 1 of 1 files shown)


docs/Encoders/svt-av1.md
```diff
@@ -1 +1,118 @@
...
+av1an ... -v " --rc 0 --crf 24 --preset 4 --input-depth 10 --tune 0" ...
+```
+
+## Variable bit-rate
+
+```
+--rc 1 --tbr
+```
+
+Variable bit-rate. Requires you to set target bit-rate `--tbr`. Recommended to use 2 passes for better quality as well, make sure to set two pass in Av1an, not SVT!
```
shssoichiro, line 39: I don't think 2-pass VBR currently works properly with av1an, given the way chunking works. You'll end up with each chunk averaging your target bitrate rather than the whole video averaging your target bitrate.

docs/Encoders/svt-av1.md
```diff
@@ -1 +1,118 @@
...
+
+VQ is subjective quality, while PSNR is an objective measurement. Most will recommend using VQ, it seems to make the image sharper as well. Default is PSNR.
+
+# Film grain
+
+```
+--film-grain --film-grain-denoise
+```
+
+Synthesize film grain! `--film-grain` can be set to 1-50 (default is 0 - off), the higher the number the stronger the effect. The default behaviour is to denoise, then add the synthesized noise. But this can remove fine detail, so it is recommended to disable the denoise stage by setting `--film-grain-denoise` to 0.
```
shssoichiro, line 78: Maybe worth noting that if your source is noisy, you can get better results by denoising it prior to encoding. (There are ffmpeg/vapoursynth denoisers that are higher quality than what any encoder provides internally.)

docs/Encoders/svt-av1.md
```diff
@@ -1 +1,118 @@
...
+... --film-grain 10 --film-grain-denoise 0 ...
+```
+
+# Input depth
+
+```
+--input-depth 10
+```
+
+You can choose a bit-depth of 8bit or 10bit. It is almost always recommended to use 10bit, even if the source is only 8bit. It is by far the most optimized, and can even fix problems inherit with 8bit.
```
shssoichiro, line 92:
```suggestion
You can choose a bit-depth of 8bit or 10bit. It is almost always recommended to use 10bit, even if the source is only 8bit. It is by far the most optimized, and can even fix problems inherent with 8bit.
```

docs/Encoders/svt-av1.md
```diff
@@ -1 +1,118 @@
...
+
+You might not want to use 10bit if fast encode/decode is more important than video quality.
+
+# Lookahead and key frames
+
+```
+--lookahead --keyint
+```
+
+These optimisation come at the cost of decoding performance, but are worth the compromise.
```
shssoichiro, line 102:
```suggestion
These optimisations come at the cost of decoding performance, but are worth the compromise.
```

shssoichiro, line 102: I don't believe lookahead has a significant decoding performance cost, although it will increase the encoder's memory usage. On the decoder side, keyint should only affect seeking performance.

docs/Encoders/svt-av1.md
hansejo: I do know 2-pass is possible in aom. Do we need to set 2 pass in the svt options instead?

docs/Encoders/svt-av1.md
hansejo: I know the ffmpeg filters. But I don't know anything about vaporsynth, what filters do they have?

docs/Encoders/svt-av1.md
shssoichiro: It "works" but it doesn't work the way you expect (with aom or svt). I think a (terrible) drawing might be the easiest way to explain it.

Running 2-pass VBR in aomenc (or any encoder) directly, you would see a bitrate distribution like this (red = keyframes):

expected-vbr


However, because of how chunking works in av1an, attempting to do 2-pass VBR in av1an will get you something like this:

av1an-vbr

Which is very different than we expect. Because each chunk averages to our requested bitrate, it means that chunks which need more bitrate will be starved of it, and chunks which need less bitrate will be bloated.

docs/Encoders/svt-av1.md
shssoichiro: Some of the most commonly used high quality denoisers for Vapoursynth include BM3D, DFTTest, SMDegrain, and KNLMeansCL. Which I'm sure ffmpeg also has some similar filters.

docs/Encoders/svt-av1.md
hansejo: Thanks for the info. I have removed the advice for the time being.

hansejo:
@shssoichiro Ready for second review.

shssoichiro (approved):
Thanks, this is good stuff!

PR #616, opened 2022-04-17
Remove data streams when encoding audio

GaveUp:
If an input file contains any data streams audio encoding will fail when trying to write a non-supported stream to the intermediary mkv.  This removes any data streams found in the input.  

```
Output {
    status: ExitStatus(
        unix_wait_status(
            256,
        ),
    ),
    stdout: "",
    stderr: "[matroska @ 0x565479976f00] Only audio, video, and subtitles are supported for Matroska.\nCould not write header for output file #0 (incorrect codec parameters ?): Invalid argument\nError initializing output stream 0:1 -- \n",
}
```

### DIFF
--- a/av1an-core/src/ffmpeg.rs
+++ b/av1an-core/src/ffmpeg.rs
@@ -170,6 +170,7 @@ pub fn encode_audio<S: AsRef<OsStr>>(
       "-map_metadata",
       "0",
       "-vn",
+      "-dn",
       "-map",
       "0",
       "-map",


av1an-core/src/ffmpeg.rs
```diff
@@ -174,6 +174,8 @@ pub fn encode_audio<S: AsRef<OsStr>>(
       "0",
       "-map",
       "-0:a",
+      "-map",
+      "-0:d",
```
shssoichiro, line 178: Could you explain what's going on here? From my knowledge of ffmpeg, this should cause the data stream to be included, right?

redzic:
I think it would be easier to do this by just adding `-dn` to the parameters like we do with `-vn`, if that also fixes it.

av1an-core/src/ffmpeg.rs
GaveUp: A `-` before the stream index means deselect.

GaveUp:
You are correct.  `-dn` makes more sense and has the same effect.

redzic (approved):
Thanks!

PR #618, opened 2022-04-18
Error if scene change frame count does not match

shssoichiro:
This way potential issues with input sources or filters can be caught sooner.


shssoichiro:
```
error: the function has a cognitive complexity of (26/25)
  --> av1an-core/src/scene_detect.rs:71:8
   |
71 | pub fn scene_detect(
   |        ^^^^^^^^^^^^
   |
```

despair

redzic:
I think we should probably error here, I'm not sure if it would ever make sense for the user to ignore this scenario

redzic (approved):

PR #619, opened 2022-04-20
Consolidate import style in rustfmt.toml

redzic:


shssoichiro (approved):
Thanks, I like this a lot.

PR #620, opened 2022-04-21
Add quotes to ffmpeg concat method

doctortheemh:
The ffmpeg concat demuxer can take quoted filenames which allows for spaces in the paths passed to ffmpeg. This aligns with the recommendation from the [wiki](https://trac.ffmpeg.org/wiki/Concatenate).


redzic:
I don't think this is necessary, since spaces are already escaped with a backwards slash. Does this fix a case that wasn't working before?

doctortheemh:
No, I think you're right. I'll do some more testing with spaces to confirm there isn't an issue.

PR #622, opened 2022-04-25
Improvements to grain synth gamma handling

shssoichiro:
- Fixes BT.1886 to use the more correct formula and gamma
- Simplify the `mid_tone` function
- Add quickcheck tests around `to_linear`/`from_linear`

### DIFF
--- a/av1an-core/src/grain.rs
+++ b/av1an-core/src/grain.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::inline_always)]
+
 use std::fs::File;
 use std::io::{BufWriter, Write};
 use std::path::Path;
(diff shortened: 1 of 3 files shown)


av1an-core/src/grain.rs
```diff
@@ -42,19 +85,27 @@ impl TransferFunction {
...
+      TransferFunction::BT1886 => {
+        // Scale to a raw cd/m^2 value
+        let luma = x * BT1886_WHITEPOINT;
+
+        // The inverse of the `to_linear` formula:
+        // `(L / α)^(1 / λ) - β = x`
+        (luma / bt1886_alpha()).powf(1.0 / BT1886_GAMMA) - bt1886_beta()
+      }
       TransferFunction::SMPTE2084 => {
+        if x == 0.0 {
```
redzic, line 97: Floating point comparisons to 0 should probably be done against an epsilon, like `x.abs() < f32::EPSILON`

av1an-core/src/grain.rs
```diff
@@ -42,19 +85,27 @@ impl TransferFunction {
...
     }
   }
 
+  #[inline(always)]
   pub fn mid_tone(self) -> f32 {
-    match self {
-      TransferFunction::BT1886 => 0.18,
-      TransferFunction::SMPTE2084 => 26. / 10000.,
-    }
+    self.to_linear(0.5)
```
redzic, line 108: This seems like it should call `from_linear` (since `from_linear` takes a value between 0 and 1), but I'm not sure. Could you explain why `to_linear` is called here?

shssoichiro: The midtone we're seeking here is the perceptual gray midtone according to the transfer function, i.e. based on this transfer function, what value with the transfer function applied appears to be "half brightness". This is calculated by doing the transfer calculation on a mathematically half luma value, i.e. `to_linear(0.5)`, and for e.g. a flat gamma curve of 2.4, results in a midtone of about `0.189`.

av1an-core/src/grain.rs
```diff
@@ -28,10 +30,51 @@ const PQ_C1: f32 = 3424. / 4096.;
...
+}
+
 impl TransferFunction {
   pub fn to_linear(self, x: f32) -> f32 {
     match self {
-      TransferFunction::BT1886 => x.powf(2.8),
+      TransferFunction::BT1886 => {
+        // The screen luminance in cd/m^2:
+        // L = α * max((x + β, 0))^λ
+        let luma = bt1886_alpha() * 0f32.max(x + bt1886_beta()).powf(BT1886_GAMMA);
```
redzic, line 73: `max(x + β, 0)` as opposed to just `x + β` inside this function appears to have the same value for all inputs, see https://www.desmos.com/calculator/1lxvcqwtny

redzic: Nevermind, I completely messed up that comparison.

redzic (approved):

av1an-core/src/grain.rs
shssoichiro: I.e. we can drop the max and this will return the same results? sounds good to me. Interesting that the formula had the `max` there considering all inputs are restricted to 0 <= x <= 1.

redzic: I guess we could still remove it then? I also noticed that the inverse of `to_linear` seemed to be derived from a version of `to_linear` without the `max`, so I guess it would make sense if we could remove it.

PR #626, opened 2022-05-04
Print a proper error message if the input file does not exist

shssoichiro:
Currently a missing input file will cause the encoder to panic later on. This change causes av1an to fail early with a clear error message.

### DIFF
--- a/av1an-cli/src/lib.rs
+++ b/av1an-cli/src/lib.rs
@@ -519,12 +519,12 @@ pub(crate) fn resolve_file_paths(path: &Path) -> anyhow::Result<Box<dyn Iterator
   // TODO: to validate file extensions
   // let valid_media_extensions = ["mkv", "mov", "mp4", "webm", "avi", "qt", "ts", "m2t", "py", "vpy"];
 
-  if path.is_file() {
-    Ok(Box::new(std::iter::once(path.to_path_buf())))
-  } else if path.is_dir() {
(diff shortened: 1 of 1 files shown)


redzic:
For some reason for me this still doesn't seem to print the correct error message when the file does not exist

shssoichiro:
flat_map was swallowing the errors from `resolve_file_paths` apparently and just removing them from the inputs array. Weird behavior, and honestly unexpected from the standard library.

redzic:
I think it has to do with how `filter_map` is used in our `read_in_dir` implementation, which apparently throws out `Err` or `None` values from the iterator, and thus we end up with an empty vector. I *think* flat_map itself is fine

redzic:
Could we just remove the `resolve_file_paths` function and replace the start of the `parse_cli` function with something like this?

```rust
  let mut input_paths = args.input;

  let mut inputs = Vec::with_capacity(input_paths.len());

  for path in input_paths.iter_mut() {
    ensure!(path.exists(), "Input path {:?} does not exist. Please ensure you typed it properly and it has not been moved.", path);

    if path.is_file() {
      inputs.push(std::mem::take(path));
    } else if path.is_dir() {
      for path in std::fs::read_dir(path.as_path())? {
        let path = path?;
        let file_type = path.file_type()?;
        if file_type.is_file() {
          inputs.push(path.path());
        }
      }
    }
  }
```

redzic (approved):
Looks good, maybe in a future PR we could remove the `Box::new` somehow to reduce allocations

PR #631, opened 2022-05-13
Implement alternate input for scene detection

BlueSwordM:
This patch came from this av1an branch made by @redzic 
https://github.com/master-of-zen/Av1an/tree/input-sc

I just refactored to make it work on master, but that's about it.

I've been using it for months(always on the same input source), and it's been working well to prevent my Vapoursynth filters from wrecking the scene detection speed, and I haven't had any problems.

We can expand this to make it work for other kind of inputs(ffmpeg) and make it work in target-quality, but otherwise, that's fine :)

### DIFF
--- a/av1an-cli/src/lib.rs
+++ b/av1an-cli/src/lib.rs
@@ -116,6 +116,10 @@ pub struct CliOpts {
   #[clap(short, parse(from_os_str), required = true)]
   pub input: Vec<PathBuf>,
 
+  /// Alternate input for scene detection (must have the same frame count)
+  #[clap(short = 'I', parse(from_os_str))]
+  pub input_sc: Option<PathBuf>,
+
   /// Video output file
   #[clap(short, parse(from_os_str))]
   pub output_file: Option<PathBuf>,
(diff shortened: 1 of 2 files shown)


shssoichiro (dismissed):
I don't see why not

redzic:
There are 2 slight issues with this at the moment.

1. If there is a different frame count, we get a panic message (something like `thread 'main' panicked at 'scenecut reported at index 22320, but there are only 144 frames', av1an-core/src/split.rs:58:5`). Ideally this error message could be improved on.
2. Since we now support multiple inputs, maybe make it work for multiple inputs somehow or maybe explicitly only allow this option to be used when a single input is being used.

BlueSwordM:
Personally, even if the tests passed, I'd wait until redzic can come in and see what he can make better :)

redzic:
So @BlueSwordM since you opened this PR, would you like to try to address the issues or do you want me to reopen it and try to work on it?

BlueSwordM:
Yeah, you can do it if you want.

I'm currently working in other stuff not related to AV1 and video stuff at all, so it would be appreciated @redzic

redzic:
Sounds good, I just wanted to ask so that the PR doesn't just sit for a long time with nothing happening. I'll try to work on this whenever I am able to make time for it.

BlueSwordM:
I see this needs to be refactored since I'm seriously restarting my encodes.

PR #632, opened 2022-05-16
Improve selection of HDR content

shssoichiro:
Currently this only applies to grain synth generation,
but the prior implementation of HDR detection only looked at the props
on the input video. Those may not necessarily be set correctly even if
the user is encoding to HDR, so we want to prioritize looking at the
encoder params, then fallback to checking the input if the user has not
set a transfer characteristic on the encoder.

### DIFF
--- a/av1an-core/src/lib.rs
+++ b/av1an-core/src/lib.rs
@@ -158,7 +158,7 @@ impl Input {
     })
   }
 
-  pub fn transfer_function(&self) -> anyhow::Result<TransferFunction> {
+  fn transfer_function(&self) -> anyhow::Result<TransferFunction> {
     const FAIL_MSG: &str = "Failed to get transfer characteristics for input video";
     Ok(match self {
       Input::VapourSynth(video) => {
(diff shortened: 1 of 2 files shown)


av1an-core/src/lib.rs
```diff
@@ -180,6 +180,31 @@ impl Input {
...
+    enc_params: &[String],
+  ) -> anyhow::Result<TransferFunction> {
+    if enc_params.iter().any(|p| {
+      let p = p.to_lowercase();
+      p == "pq" || p.ends_with("=pq") || p.ends_with("smpte2084")
+    }) {
+      return Ok(TransferFunction::SMPTE2084);
+    }
+    if enc_params.iter().any(|p| {
+      let p = p.to_lowercase();
```
redzic, line 194: My only (extremely minor) nitpick is that these 2 `to_lowercase` function calls could instead be `to_ascii_lowercase`, which are theoretically a lot faster because they don't have to load values from a huge utf8 lookup table.

shssoichiro: Technically true. It won't affect overall performance as this function is called at most twice per encode, but since it's still in PR might as well make it the better version.

redzic (approved):

PR #633, opened 2022-05-21
Use std::available_parallelism instead of num_cpus

shssoichiro:
The [available_parallelism](https://doc.rust-lang.org/std/thread/fn.available_parallelism.html) function was made stable in Rust 1.59. This changeset makes use of it instead of the `num_cpus` crate. We still have transitive dependencies that use `num_cpus`, but we should still prefer to use stdlib code when possible.

This changeset also explicitly sets the minimum Rust version to 1.59. This was already the case, as it was required by Edition 2021, but now Cargo will show a helpful error message to users on an old-ish version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -196,7 +196,6 @@ dependencies = [
  "log",
  "memchr",
  "nom",
- "num_cpus",
  "once_cell",
  "parking_lot",
  "paste",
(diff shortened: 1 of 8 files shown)


redzic (approved):
Hopefully we can eventually switch all the dependencies away from num_cpus as well too.

PR #634, opened 2022-06-05
Don't panic if ffmpeg fails to count frames

Y0ba:
Sometimes with broken input `num_frames` fails to count frames and `.unwrap()` panics. Unfortunately since it panics in threaded code, this panic does not stop app from running and instead thread just hangs forever and tui shows some weird visual glitches.

### DIFF
--- a/av1an-core/src/settings.rs
+++ b/av1an-core/src/settings.rs
@@ -427,20 +427,32 @@ impl EncodeArgs {
     }
 
     if current_pass == passes {
-      let encoded_frames = num_frames(chunk.output().as_ref()).unwrap();
+      let encoded_frames = num_frames(chunk.output().as_ref());
 
-      if encoded_frames != chunk.frames {
+      let err_str = match encoded_frames {
+        Ok(encoded_frames) if encoded_frames != chunk.frames => {
+          Some(format!(
+            "FRAME MISMATCH: chunk {}: {}/{} (actual/expected frames)",
+            chunk.index, encoded_frames, chunk.frames
+          ))
+        },
+        Err(error) => {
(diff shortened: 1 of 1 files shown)


redzic:
Looks like you need to run `cargo fmt`, that's why the CI is failing. I think this looks OK for now, but maybe in the future we want to retry the frame count itself instead of restarting the entire chunk.

Y0ba:
In my case retrying didn't help at all (it failed constantly). Maybe it will be better to catch all panics in children threads and exit from application?

It's possible to do that with code like this (in `main.rs`):
```rust
use av1an_cli::run;
use std::panic;
use std::process;

fn main() -> anyhow::Result<()> {
  let orig_hook = panic::take_hook();
  panic::set_hook(Box::new(move |panic_info| {
    orig_hook(panic_info);
    process::exit(1);
  }));
  run()
}
```

redzic:
Just out of curiosity, what are the circumstances in which you can reproduce this frame count failure to happen? I ask because maybe there's another solution to that particular cause of this issue.

Your suggestion about catching panics in child threads seems fine to me. I think it'd be better if we do that in a followup PR though.

redzic (approved):
Thanks! Definitely an improvement over the current code.

Y0ba:
> Just out of curiosity, what are the circumstances in which you can reproduce this frame count failure to happen?

`lsmash` failed to open `.VOB`-file. Either using `ffms2` or remuxing into `.mkv` fixed the error.

redzic:
Yeah, in that case retrying wouldn't help at all. Well at least this error doesn't seem to happen in situations where it shouldn't.

Issue #240, opened 2021-03-28
Improve logging for first pass failures


cromefire:
For some reason it seems like some first passes may be missing / failed (there's sadly isn't any indication which command spit out those errors), but there isn't anything useful logged, maybe some more error logging would be good.

Version: 6c81ea1

<details>
<summary>Log</summary>

```
[20:49:29] Av1an Started
[20:49:29] Command:
[20:49:29] .local/bin/av1an --encoder aom -p 2 -i input.mkv -o output.mkv -fmt yuv420p10le -r --keep -w 6 -cm hybrid -v  --height=2160 --width=3840 --good --bit-depth=10 --end-usage=cq --cq-level=30 --input-bit-depth=10 --threads=8 --cpu-used=6 --denoise-noise-level=5 
[20:49:29] Starting scene detection Threshold: 35, Min_scene_length: 60
[20:49:29] Starting PySceneDetect:
[20:49:29] Threshold: 35
[20:49:29] Min scene length: 60
[20:49:29] Is Vapoursynth input: False
[20:49:30] Found scenes: 0
[21:07:52] Applying extra splits
[21:07:52] Split distance: 240
[21:07:52] New splits:689
[21:25:45] Split Video
[21:42:04] Split Done
[21:42:04] Audio processing
[21:42:04] Params: -c:a copy
[21:50:53] Enc: 1, 240 fr
[21:50:53] Enc: 2, 240 fr
[21:50:53] Enc: 4, 240 fr
[21:50:53] Enc: 5, 240 fr
[21:50:53] Enc: 7, 240 fr
[21:50:53] Enc: 8, 240 fr
[22:02:09] Done: 5 Fr: 240/240
[22:02:09] Fps: 0.3551 Time: 675.84 sec.
[22:02:09] Enc: 9, 240 fr
[22:02:50] Done: 8 Fr: 240/240
[22:02:50] Fps: 0.3346 Time: 717.19 sec.
[22:02:50] Enc: 11, 240 fr
[22:03:00] Done: 2 Fr: 240/240
[22:03:00] Fps: 0.3303 Time: 726.71 sec.
[22:03:00] Enc: 12, 240 fr
[22:03:24] Done: 7 Fr: 240/240
[22:03:24] Fps: 0.3198 Time: 750.36 sec.
[22:03:24] Enc: 14, 240 fr
[22:04:50] Done: 1 Fr: 240/240
[22:04:50] Fps: 0.287 Time: 836.25 sec.
[...]

PR #636, opened 2022-06-05
Fix x265 2-pass params

Y0ba:
Seems like it was copypasted from `x264`, but in `x265` some options are different. Plus added [`--analysis-reuse-file`](https://x265.readthedocs.io/en/master/cli.html#cmdoption-analysis-reuse-file) option which is used for [`--multi-pass-opt-analysis`](https://x265.readthedocs.io/en/master/cli.html#cmdoption-multi-pass-opt-analysis) and [`--multi-pass-opt-distortion`](https://x265.readthedocs.io/en/master/cli.html#cmdoption-multi-pass-opt-distortion) and ignored if neither specified.

### DIFF
--- a/av1an-core/src/encoder.rs
+++ b/av1an-core/src/encoder.rs
@@ -208,16 +208,23 @@ impl Encoder {
       Self::x265 => chain!(
         into_array![
           "x265",
-          "--stitchable",
+          "--repeat-headers",
           "--log-level",
           "error",
           "--pass",
           "1",
-          "--demuxer",
-          "y4m",
+          "--y4m",
         ],
         params,
-        into_array!["--stats", format!("{}.log", fpf), "-", "-o", NULL]
+        into_array![
+          "--stats",
+          format!("{}.log", fpf),
+          "--analysis-reuse-file",
+          format!("{}_analysis.dat", fpf),
+          "-",
+          "-o",
+          NULL
+        ]
       )
       .collect(),
     }
@@ -283,16 +290,23 @@ impl Encoder {
       Self::x265 => chain!(
         into_array![
           "x265",
-          "--stitchable",
+          "--repeat-headers",
           "--log-level",
           "error",
           "--pass",
           "2",
-          "--demuxer",
-          "y4m",
+          "--y4m",
         ],
         params,
-        into_array!["--stats", format!("{}.log", fpf), "-", "-o", output]
+        into_array![
+          "--stats",
+          format!("{}.log", fpf),
+          "--analysis-reuse-file",
+          format!("{}_analysis.dat", fpf),
+          "-",
+          "-o",
+          output
+        ]
       )
       .collect(),
     }
--- a/av1an-core/src/encoder.rs
+++ b/av1an-core/src/encoder.rs
@@ -409,7 +409,16 @@ impl Encoder {
         }
       }
       Encoder::x264 => into_vec!["--preset", "slow", "--crf", "25"],
-      Encoder::x265 => into_vec!["-p", "slow", "--crf", "25", "-D", "10"],
+      Encoder::x265 => into_vec![
+        "-p",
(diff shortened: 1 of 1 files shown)


redzic:
Thanks again! Looks like it needs a rebase though.

Y0ba:
Rebased.

redzic:
Hmm, it looks there's 2 issues now.
1. The default x265 options (specified by av1an) don't work with 2-pass. Something about CRF mode being incompatible with 2-pass. Perhaps we could change the default x265 options if 2-pass mode is enabled.
2. The frame counting seems to count the first pass as well, which is not supposed to happen. This causes the progress bar to go past 100%. I'll see if I can figure out the cause of this.

redzic:
The cause of the second issue seems to be that the parsing code relies on returning `None` based purely on the output, without prior knowledge of the encoding pass. This is not possible with the output of x265 in 2-pass mode. The "real" fix would be to not actually parse the first pass progress. I'll try to open a PR that does this, so that this PR can work with minimal friction.

redzic:
Rebasing should fix the second issue.

Y0ba:
Error in the first issue is:
`Constant rate-factor is incompatible with 2pass without vbv-maxrate in the previous pass`
It's possible to specify `--vbv-bufsize 35000 --vbv-maxrate 35000`. 35000 Should be safe for all hardware decoders, and more than enough for default `--crf 25` which probably will never be capped.

Y0ba:
Or it's possible to just use `--level-idc 5.1` (or `4.1`, but it doesn't support 4k). Then `vbv` will be automatically adjusted according to specified level.

redzic:
Specifying `--level-idc 5.1` seems fine to me. The function `get_default_arguments` would have to be updated to take in the number of passes, and of course the function body as well too.

Y0ba:
--level-idc 5.1 can be specified for single pass too. It won't change much except making video incompatible with some old hardware decoders. In my tests by default x265 encodes videos with 4.1 level.

redzic:
I see, thanks. In that case that makes things even more simple. Not sure if we want to detect which profile to use based on the resolution, but I'm fine with simply adding `--level-idc 5.1` to the default x265 parameters.

Y0ba:
I think `4.1` might be a good default. `4k` almost always comes with hdr and it requires specifying custom flags (at least `--transfer`, `--colormatrix` and `--colorprim`). I'm not sure anyone will use default settings for 4k encodes.

redzic:
That seems OK, but for some reason I seem to get infinite stalling with 4k SDR with `--level-idc 4.1`, no error message unless I specify `-p1`. I would prefer if the default settings do at least work or at the very least display a proper error message.

Y0ba:
Yeah, in 2 pass encodes if encoder fails in 1 pass, then it just stalls. That's #240 I I think.

redzic:
That shouldn't be the case anymore as of the Rust version. The issue you linked was opened when the project was still written in Python. If I add an invalid option to the first pass on purpose (like, in the `compose_1_2_pass` function itself), the error is correctly reported:

```
ERROR [av1an_core::broker] [chunk 1] encoder failed 3 times, shutting down worker
ERROR [av1an_core::broker] [chunk 1] encoder crashed: exit status: 1
stdout:

stderr:
        Error: Unrecognized option --invalid-option

        Usage: aomenc <options> -o dst_filename src_filename
        Use --help to see the full list of options.


source pipe stderr:

ffmpeg pipe stderr:
```

The strange thing is that the x265 processes still exist, it just stalls infinitely.

Y0ba:
OK, I've added 5.0 level which is a nice middle-ground since it supports 4k (up to 32fps) and can be hw-decoded even in my old tablet made in 2016.

redzic (approved):
Seems fine to me. Thanks!

Issue #530, opened 2022-01-14
Auto-compile Windows binaries as part of CI pipeline


shssoichiro:
At first maybe we start by only doing it when we tag a release? We'd want to start releasing more frequently, though, I think, if that happens. But the auto-builds would also enable us to release more often.

FreezyLemon:
Do you have a preference on how this should work?

I could imagine having the github action create a draft release for you to add a description to and then publish (exe auto-included), but there are other possibilities. Maybe pre-releases in a 'nightly' style on every push, or fully automated releases every time a new tag is pushed.

shssoichiro:
> Maybe pre-releases in a 'nightly' style on every push

This is essentially what I was thinking with this. The PR you put together looks nice. 🙂

PR #638, opened 2022-06-15
Create workflow for Windows x64 builds

FreezyLemon:
Starting point for #530.
Builds an x64 executable and uploads it as a workflow artifact ([example run](https://github.com/FreezyLemon/Av1an/actions/runs/2504902829))

The [exact](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows) workflow trigger is still open for discussion. And maybe there's some way to automatically add the artifact to a release? Haven't looked into it yet, this is just the build.

Dependencies:
- Visual C++ Build Tools: pre-installed
- Python 3.10: pre-installed + setup by action for proper env variables
- NASM: Installed by action
- FFmpeg: Manual download (gyan.dev), cached via file hash
- VapourSynth: Manual download, cached via release version

Cargo packages are also cached. NASM could also be cached, but I didn't see it as a high priority.

Happy to hear your thoughts

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -1,0 +1,90 @@
+name: Build Av1an (Windows x64)
+on: workflow_dispatch
+
+jobs:
+  build:
+    runs-on: windows-2022
+
+    env:
+      vsynth_ver: R58
+      vsynth_path: C:\Program Files\Vapoursynth
+
+    steps:
+      - name: Python 3.10 setup
+        uses: actions/setup-python@v4
+        with:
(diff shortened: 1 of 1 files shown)


redzic:
Thanks! I'm not sure why the clippy tests are failing now and not before, but I've opened #639 to fix it. Rebasing after that should in theory fix the CI failures.

redzic (approved):
Looks good to me as a starting point, thanks! Perhaps we could figure out how to automatically attach build artifacts to releases in a future PR.

PR #640, opened 2022-06-22
Change progress bar characters

redzic:
This change brings the look of the progress bar to be more in line with the previous Python version of av1an, which had a solid progress bar, and it looks much smoother in general now as well.

Before:

image

After:

image


p7r0x7:
~~I see this as a regression.~~

PR #641, opened 2022-06-25
vmaf_percentile to 1, default vmaf rate to 1

master-of-zen:
I think target quality need a lot of work and improvement.

* vmaf_percentile to 1, 
* default vmaf rate to 1

### DIFF
--- a/av1an-cli/src/lib.rs
+++ b/av1an-cli/src/lib.rs
@@ -462,7 +462,7 @@ pub struct CliOpts {
   pub probes: u32,
 
   /// Framerate for probes, 1 - original
-  #[clap(long, default_value_t = 4, help_heading = "TARGET QUALITY")]
+  #[clap(long, default_value_t = 1, help_heading = "TARGET QUALITY")]
   pub probing_rate: u32,
 
   /// Use encoding settings for probes specified by --video-params rather than faster, less accurate settings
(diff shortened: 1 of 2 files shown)


BlueSwordM:
Denied. We should put it to vmaf_percentile to 5% (0.05) and the default vmaf subsampling rate to 1/2.

In fact, we should introduced VMAF percentile controls like I tried to do here, but then stopped:
https://github.com/BlueSwordM/Av1an/tree/av1an_tq_percentile

master-of-zen:
I'm certain that those are needed changes
First is that GOP structure, and pyramid changes if you have VMAF rate >1, which impacts probes accuracy, in quite unpredictable way, even more on anything that have motion.
Second is that I have seen in many instances where results were quite suboptimal because of high percentile, the idea behind having it at 25 initially was to capture those frames that are worse and have disproportional impact to perceived quality.
On my revisiting of that decision, and looking into specific instances where catching bad frames  25 percentiles were not enough.  It's simply not as aggressive as it should be, I have been noticing quality difference with less than 5 frames that had bad results, and they were not reflected by 25 percentile.

redzic (approved):
In addition to updating `adapt_probing_rate`, the actual default value needed to be updated as well. Other than that, I think these changes make sense as a default, and it also makes it more consistent with the default scene detection settings, in that they are the highest accuracy/quality by default. Other issues with target quality can be addressed in future PRs.

Issue #642, opened 2022-06-27
cargo chef fails with `profile `ci` is missing an `inherits` directive`


Extarys:
With latest pull from av1an/master, I can't no longer build the docker image.

```
Step 13/27 : RUN cargo chef cook --release --recipe-path recipe.json
[...]
  Downloaded rayon-core v1.9.1
  Downloaded pin-project-lite v0.2.8
  Downloaded num_threads v0.1.5
  Downloaded crossbeam-utils v0.8.8
  Downloaded bindgen v0.59.2
  Downloaded cc v1.0.73
  Downloaded bytes v1.1.0
  Downloaded arrayvec v0.7.2
  Downloaded unicode-normalization v0.1.19
  Downloaded idna v0.2.3
  Downloaded indexmap v1.8.0
  Downloaded git2 v0.14.2
  Downloaded flexi_logger v0.22.3
  Downloaded ffmpeg-sys-next v5.0.1
  Downloaded anyhow v1.0.56
  Downloaded libc v0.2.121
  Downloaded libgit2-sys v0.13.2+1.4.2
  Downloaded libz-sys v1.1.5
  Downloaded plotters v0.3.1
error: profile `ci` is missing an `inherits` directive (`inherits` is required for all profiles except `dev` or `release`)
thread 'main' panicked at 'Exited with status code: 101', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-chef-0.1.36/src/recipe.rs:148:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
The command '/bin/sh -c cargo chef cook --release --recipe-path recipe.json' returned a non-zero code: 101
```

redzic:
Possibly a bug in cargo-chef? Removing the `ci` profile from Cargo.toml fixes the docker build for me, but unfortunately I don't think that's feasible since the scene detection on the debug profile is unbearably slow, and we wanted to still enable overflow checks to catch errors in CI.

Also strange because the `ci` profile isn't actually missing an `inherits` directive.

FreezyLemon:
Update: This has been [fixed](https://github.com/LukeMathWalker/cargo-chef/pull/129#issuecomment-1172378703) upstream. The issue was a bug in a dependency of cargo-chef. Reverting #643 should work. (Oh but please keep the fixed `pacman -Syu` at the top)

Extarys:
Should be fixed.

PR #643, opened 2022-06-28
Temporarily disable docker caching via cargo-chef

redzic:
Unfortunately cargo-chef no longer seems to work for this repository; it complains about a missing 'inherits' directive for the CI profile, even though it is not actually missing. Additionally, downgrading to previous versions also somehow causes errors.

Removing the CI profile fixes the docker build, however, since the scene detection is too slow for the CI in the debug build, and since we still would like overflow checking and debug *assertions* (and hence can't use the release profile), removing the CI profile is not a feasible workaround for this issue.

Therefore, the only solution for now seems to be to disable cargo-chef altogether, as being able to build via docker is preferable to being unable to do so at all, even if it is not as fast as it could be with docker layer caching.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM archlinux:base-devel AS base
 
-RUN pacman -Syy --noconfirm
+RUN pacman -Syyu --noconfirm
 
 # Install dependancies needed by all steps including runtime step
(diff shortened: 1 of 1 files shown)


master-of-zen (approved):

Extarys:
For future references, one can look at issue https://github.com/master-of-zen/Av1an/issues/642

PR #646, opened 2022-07-04
Automate Windows binary releases on changes to master

FreezyLemon:
Second part of #530.

First commit has some general improvements. The improved caching cuts the build time by around half.

Second commit kicks off the workflow on every push to `master` and automatically creates a prerelease with the Windows binary. This release is always called `latest` and can only exist once. This means the previous `latest` release will be deleted and then re-created with the new Windows binary. The latest release can always be found under the same URL (example https://github.com/FreezyLemon/Av1an/releases/tag/latest).

This does not cover full, tagged releases. But those can be made by downloading the last `av1an.exe` and manually attaching it to the release, which makes it easier, I think.

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -1,5 +1,8 @@
 name: Build Av1an (Windows x64)
-on: workflow_dispatch
+on:
+  push:
+    branches:
+      - 'master'
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


redzic (approved):
Looks good, thanks! This should make downloading newer versions of av1an significantly easier.

PR #649, opened 2022-07-08
Update readme

FreezyLemon:
After seeing that the README was pretty outdated, I decided to try and update it a bit. Some things I'm not 100% about:

- The first few sentences. These are pretty important as far as READMEs go, so feel free to edit/propose stuff
- Saying `lsmash` is the recommended choice. This just seemed like the generally preferred choice from what I read in Discord 
- I think "use `av1an --help` for a CLI reference" is easier to maintain than listing every option in the readme
- Where to put the image of the 96-worker encode

Feel free to give feedback on these things and in general.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,7 @@ TEMP STUFF
 
 Av1an is written in Rust and can be used on Linux, macOS and Windows. It is highly configurable but tries to set good default values to make it easier to use.
 
-### Supported encoders
-
-- AV1: aomenc, rav1e, SVT-AV1
-- H.264/AVC: x264
-- H.265/HEVC: x265
-- VP8, VP9: VPX
+Binary releases for Windows are also available from this repository's [releases page](https://github.com/master-of-zen/Av1an/releases).
 
 ---
 
@@ -24,6 +19,7 @@ Av1an is a video encoding framework for modern encoders. It can increase your en
 ## Table of Contents
 
 - [Installation](#installation)
+- [Supported encoders](#supported-encoders)
 - [Usage](#usage)
 - [Configuration](#configuration)
 - [Building](#building-av1an)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,261 +1,79 @@
...
+- [Target Quality](/docs/TargetQuality.md) mode, using [VMAF](https://github.com/Netflix/vmaf) to automatically set encoder options to achieve the wanted visual quality
+- Simple and clean console interface
+- Convenient Docker images available
+- Cross-platform application written in Rust
 
-Example with default parameters:
+## Supported encoders
 
-    av1an -i input
+At least one encoder is required to use Av1an. Install any of these that you wish to use.
```
redzic, line 23: Might be good to clarify the standalone binaries for each encoder that are needed for av1an to be able to use them? For example, vpxenc is needed for libvpx, and it's possible to have libvpx installed without the vpxenc binary (and this does happen in practice for some package managers).

README.md
```diff
@@ -1,261 +1,79 @@
...
-To specify a different directory to use you would replace $(pwd) with the directory
-
-```bash
-docker run --privileged -v "/c/Users/masterofzen/Videos":/videos --user $(id -u):$(id -g) -it --rm masterofzen/av1an:latest -i S01E01.mkv {options}
+```sh
+av1an -i input.vpy -v "--cpu-used=3 --end-usage=q --cq-level=30 --threads=8" -w 10 --target-quality 95 -a "-c:a libopus -ac 2 -b:a 192k" -l my_log -o output.mkv
 ```
 
-The --user flag is required on linux to avoid permission issues with the docker container not being able to write to the location, if you get permission issues ensure your user has access to the folder that you are using to encode.
+To check all available options for your version of Av1an use `av1an -h`.
```
redzic, line 73: We should probably suggest to use `av1an --help`, since the short version of it, `-h`, prints a briefer description of the options.

redzic:
Overall I think this looks very good. I do wonder if we could also add some instructions for how to install lsmash and/or ffms2 on Windows. Basically the user just needs to put some DLL files into their vapoursynth plugins directory, but most users probably wouldn't know that in addition to knowing where to find the DLLs.

README.md
```diff
@@ -1,261 +1,79 @@
...
-</h1>
+# Av1an
 
-<h2 align="center">A cross-platform framework to streamline video encoding</h2>
+[![Discord server](https://discordapp.com/api/guilds/696849974230515794/embed.png)](https://discord.gg/Ar8MvJh)
+[![CI tests](https://github.com/master-of-zen/Av1an/actions/workflows/tests.yml/badge.svg)](https://github.com/master-of-zen/Av1an/actions/workflows/tests.yml)
+[![](https://img.shields.io/crates/v/av1an.svg)](https://crates.io/crates/av1an)
 
-![alt text](https://cdn.discordapp.com/attachments/804148977347330048/928879953825640458/av1an_preview.jpg)
+Av1an is a video encoding framework for modern encoders. It can increase your encoding efficiency by running multiple encoder processes in parallel. This can improve CPU usage and increases the speed of some AV1 encoders dramatically.
```
redzic, line 7: I might emphasize here that the reason it increases speed so much in some cases is because of poor multi-threading in the encoders themselves. I think I would also change "some AV1 encoders" to maybe just "some encoders", since libvpx also gets a similarly large speedup.

README.md
```diff
@@ -1,261 +1,79 @@
...
 
-</h4>
-<h2 align="center">Easy, Fast, Efficient and Feature-Rich</h2>
+## Features
 
-### <center>An easy way to start using AV1, HEVC/H.265, AVC/H.264, VP9, and VP8 encoders.<br> AOM, RAV1E, SVT-AV1, VPX, x265, x264 are supported.</center>
+- Vastly improved encoding speed for some encoders
+- Cancel and resume encoding without loss of progress
+- [VapourSynth](http://www.vapoursynth.com) script support
+- [Target Quality](/docs/TargetQuality.md) mode, using [VMAF](https://github.com/Netflix/vmaf) to automatically set encoder options to achieve the wanted visual quality
```
redzic, line 16: Minor nitpick, but I would probably rephase "wanted visual quality" as "desired video quality"

README.md
FreezyLemon: Sure, I replaced the lib names with the encoder names (libvpx -> vpxenc) and put a note below the list.

README.md
FreezyLemon: Good catch, I didn't know about vpxenc. Let me know if you like the new wording.

jangrewe:
I would actually prefer to have the full `--help` output in the README, as an easy reference without having the run the command and scroll back and forth in the console all the time. It doesn't need to be nicely formatted, so putting the raw output into a code block would be good enough.

FreezyLemon:
@jangrewe I don't mind including the --help output somewhere, and spending some time on formatting isn't really an issue either. The main reason that I removed the old "parameter block" from the readme was the lack of maintenance.

There was a lot of missing and some wrong information, so I thought it would be better to just tell people to use `av1an --help` instead of referencing the outdated information.

It's no real problem though, if you prefer a parameter reference that doesn't required running av1an then I can include it.

FreezyLemon:
Well, I think with the last commit I'm pretty much done for now. Feels a bit bad to leave the image at the bottom, but it doesn't really fit anywhere else.

master-of-zen:
I think preview image should be on top

PR #653, opened 2022-07-22
Migrate grain synth code to external crate and allow chroma grain synth

shssoichiro:
Migrates grain parsing and generation code to [external crate](https://github.com/rust-av/av1-grain) (which is shared with [rav1e](https://github.com/xiph/rav1e/pull/2966)).

Also adds `--chroma-noise` option to enable chroma grain synth.

This implementation is simplistic in that it only enables chroma scaling
from luma noise based on the `--photon-noise` strength. However, it's
100% more chroma noise capability than av1an had previously.


master-of-zen (approved):
Great work :1st_place_medal:

PR #655, opened 2022-07-30
Allow generating photon noise tables for rav1e

shssoichiro:
Note that the `--photon-noise-table` flag in rav1e is currently behind the `unstable` feature, so rav1e must be compiled with `--features unstable` for this to be supported.


av1an-core/src/scenes.rs
```diff
@@ -336,6 +336,23 @@ fn validate_zones_args() {
     .contains(&"--aq-mode=1".to_owned()));
 }
 
+#[test]
+fn validate_rav1e_zone_with_photon_noise() {
+  let input = "45 729 rav1e --speed 6 --photon-noise 4";
```
CartoonFan, line 341: I'm not an expert, but the tests seem to suggest "rav1e reset" instead of just "rav1e"?

CartoonFan (approved):
The descriptive changes look fine. I don't know anything about the actual code, though.

BlueSwordM (approved):
Everything seems good, although it should be noted that the only way to get grain synth inputs to work with rav1e are to build it with "unstable" features.

master-of-zen (approved):
Amazing:) :+1: :1st_place_medal:

PR #656, opened 2022-08-03
Fix CI Windows builds

FreezyLemon:
After FFmpeg release 5.1, gyan.dev updated their "latest" release from 5.0.1 to 5.1, which currently breaks the build (`ffmpeg-next` and `ffmpeg-sys-next` don't support 5.1 yet). To circumvent this in the future, I've fixed the version to 5.0.1 and switched to the GitHub mirror (older releases get removed from gyan.dev at some point).

This requires removing the cache functionality, because there's no way to get a file hash from GitHub release assets. This doesn't affect the build time much though.

I've [tested this on my fork already, and it works without a problem.](https://github.com/FreezyLemon/Av1an/actions/runs/2790530028)

Ignore the 2 commits, I'm just too lazy to squash them manually :D

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: FFmpeg hashsum
         run: |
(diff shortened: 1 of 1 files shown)


FreezyLemon:
Ok, this can be merged after #658.

redzic (approved):

PR #661, opened 2022-08-31
Fix extra_splits when fps is NaN

redzic:
Sometimes, `avg_frame_rate` returns NaN, which causes a divide-by-zero error in `extra_splits`. This PR fixes the behavior.

- Also fix other compiler warning.


master-of-zen (approved):
👍

Issue #650, opened 2022-07-11
Unable to change video fps


KyleUltimate:
I wanna change my video fps without slowing down, essentially just dropping frames.
base codec is vp9.
I tried using ffmpeg_filter_args to do. `-r 30`, the encoder complains about frame mismatch
![image](https://user-images.githubusercontent.com/68589851/178178057-c30d9dba-910a-4e74-93c6-2bbae87400a6.png)
Then I tried `-filter:v fps=fps=30`, same exact error.
Then I tried using vp9's fps filter, `--fps=30/1`, it works this time, but it slows down the footage by 50%(cause my original video is 60fps)
Even by combining vp9 fps filter + ffmpeg filter, it all comes down to the same problem.

silverbacknet:
That's not something Av1an is set up to do, since it completely changes its expectations. What you can do instead is, load the video via a VapourSynth script that performs the decimation before it reaches Av1an instead.

redzic:
Yeah, pretty much just echoing what @silverbacknet said here.

`--ffmpeg` filters each chunk separately, and the filtering is not set up in a way such that it expects filters to change the number of frames in a chunk, i.e. it does not filter the entire video first and then encode it.

A vapoursynth that changes the video fps should work though.

KyleUltimate:
Oh ok. I assume this can do it?
```python
import vapoursynth as vs
import functools
import math

target_fps_num = 30
target_fps_den = 1
core = vs.core
source_clip = core.ffms2.Source(source='abr213hrH2.mkv')
def frame_adjuster(n, clip, target_fps_num, target_fps_den):
    real_n = math.floor(n / (target_fps_num / target_fps_den * clip.fps_den / clip.fps_num))
    one_frame_clip = clip[real_n] * (len(clip) + 100)
    return one_frame_clip

attribute_clip = core.std.BlankClip(source_clip, length=math.floor(len(source_clip) * target_fps_num / target_fps_den * source_clip.fps_den / source_clip.fps_num), fpsnum=target_fps_num, fpsden=target_fps_den)
adjusted_clip = core.std.FrameEval(attribute_clip, functools.partial(frame_adjuster, clip=source_clip, target_fps_num=target_fps_num, target_fps_den=target_fps_den))
adjusted_clip.set_output()
```

lastrosade:
Changing the framerate from vapoursynth does in fact work.

Tested with

```python
import vapoursynth as vs
import mvsfunc
import lasfunc

[...]

PR #671, opened 2022-09-30
Add additional flags, options

woot000:
This commit adds:

- Never overwrite ```-n``` flag, acts as the opposite of the overwrite ```-y``` flag
- Scaler selector ```--scaler``` flag, allows the user to choose any scaler available within ffmpeg, including defining the width of the lanczos scaler by typing lanczos1 through lanczos9 (could go higher but there's not really a point to have width values above 6)
- Extra splits in seconds ```--extra-splits-sec``` flag, allows the user to define the maximum scene length in seconds (frames takes precedent if both are defined because the clap crate can't make these options conflict with each other, for some reason)
- Ignore frame mismatch warning ```--ignore-frame-mismatch``` flag, allows the user to ignore any frame mismatch warning in case an ffmpeg filter changes the frame count, or if the input file is badly encoded (acts as a workaround for #650)
- Additional option "inputres" for ```--vmaf-res```, sets the VMAF resolution to be the same as the input video's resolution if ```--vmaf-res inputres``` is used
- Different progress bar characters for Windows, all the characters display correctly on the default Windows command prompt


mqudsi:
I've rebased this on top of master here: https://github.com/mqudsi/Av1an/tree/pr671

woot000:
Rebased again (thanks @mqudsi) with some misc. minor changes, a typo correction, and added documentation to CLI.md

shssoichiro (approved):

PR #682, opened 2022-11-14
Temporarily disable steady ticks on progress bars

shssoichiro:
This fixes an issue where the fps and eta would not update until the very end of the encode.

See also console-rs/indicatif#493

### DIFF
--- a/av1an-core/src/progress_bar.rs
+++ b/av1an-core/src/progress_bar.rs
@@ -1,5 +1,4 @@
 use std::fmt::Write;
-use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


shssoichiro:
It seems like there may be further weirdness going on. The estimated fps and eta are now varying wildly from one second to the next.

shssoichiro:
It looks like this bug and the issue with the fps and eta values wildly fluctuating are two separate, unrelated issues, so I'm going to push this forward as its own fix.

master-of-zen (approved):

PR #684, opened 2022-11-16
Fix stuttering FPS and ETA on progress bar

shssoichiro:
This also allows us to reenable steady tick.
I believe steady tick never worked as expected before, but given that it is working now, the tick duration of 100ms means that rendering the progress bar at 60hz is overkill, and we are fine with the default 20hz (which is arguably still overkill).


master-of-zen (approved):
:+1:

PR #688, opened 2022-12-06
Improve verbose progress for several encoders

shssoichiro:
By passing the frame count for the chunk to the encoder, we can allow the encoder to show an improved progress bar for each chunk in verbose mode.


master-of-zen (approved):
Amazing, as usual) :+1:

PR #703, opened 2023-01-13
Update to clap 4

shssoichiro:
This does somewhat change the appearance of the help text, but the order of options remains the same, headings have been changed to title case to match the new convention of clap 4, the full list of possible values is now populated automatically from the enum but is only displayed under `--help` (full help), among other changes to accommodate how clap 4 wants things.


redzic (approved):
Thanks!

PR #714, opened 2023-01-27
Refactor Chunk to be self-contained after generation

shssoichiro:
Previously values such as photon noise, end frame, video params, and encoder would be stored outside the chunk, and the broker would need to reference additional values such as zone overrides during encoding itself. With this change, each Chunk has all of its final values stored within itself, and will not need to be modified after it is generated.


master-of-zen:
@shssoichiro Thank you a lot) that will make things easier

master-of-zen (approved):

shssoichiro:
I don't know what's going on with the SonarCloud thing

master-of-zen:
It doesn't like docker file, for whatever reason

PR #715, opened 2023-01-30
Fix panic in parse_aom_vpx_frames_sse41()

redzic:
On some rare inputs (that can happen in real world scenarios), this function panics. This PR fixes that issue. I have also fuzzed this function and the fallback `parse_aom_vpx_frames` and no other obvious panics or crashes seem to happen in debug mode, at least on the randomized inputs used in fuzzing.


master-of-zen:
:+1:

master-of-zen (approved):

PR #717, opened 2023-02-02
Fix building on some non-x64 platforms

shssoichiro:


shssoichiro:
I'm going to note that this is untested because I don't have a machine to test it on. Platforms I'm aware of that are producing the issue are aarch64 and Windows+msys. If someone could pull this and give it a build to verify, that would be great.

shssoichiro:
I've asked multiple people in Discord who have reported this issue to please test it and it seems nobody is going to. Let's just merge it and hope it works. Essentially I reverted a set of changes in the upstream ffmpeg package that fixed some clippy lints, but those clippy lints were removing necessary casts on some platforms.

redzic (approved):
Sounds good

PR #736, opened 2023-03-23
Fix estimated fps when using --resume flag

### LINKED ISSUE #726: --resume flag gives incorrect fps estimation
If a video already has frames encoded and --resume is used, the fps estimate is calculated as being way higher than it actually is
image

shssoichiro:
Fixes #726


shssoichiro:
@master-of-zen Do you think this is good to merge? I know we talked about refactoring progress in general, but for now I think it would be good to fix this bug as a stopgap.

PR #739, opened 2023-03-28
Remove tpl

master-of-zen:
Removes tpl work around and use chunk_queue size for total chunks


av1an-core/src/broker.rs
```diff
@@ -246,17 +241,8 @@ impl<'a> Broker<'a> {
           warn!("Encoder failed (on chunk {}):\n{}", chunk.index, e);
 
           if encoder == Encoder::aom
```
shssoichiro, line 243: This if is now empty, so we can remove the whole thing.

shssoichiro (approved):
Looks like rustfmt is complaining, otherwise looks good to me :+1:

Issue #733, opened 2023-03-17
Docker image 'latest' outdated (still running 0.1.0)


rsnitsch:
The tag 'latest' has not been updated for 2 years. It is still running av1an 0.1.0:

```
docker run --privileged -v "$(pwd):/videos" --user $(id -u):$(id -g) -it --rm masterofzen/av1an:latest --version
av1an 0.1.0
```

I only discovered this by excessive caution (double- and triple-checking everything due to the weird FPS issue, see #732)...

Now that I checked the docs again I see that this is mentioned in the docs (it says "(outdated) masterofzen/av1an:latest for the latest stable python release"), however (!), the example commands presented in the docs are still using the latest tag. So that's why I originally overlooked the 'outdated' part...

In general I think you should stick to the Docker convention of 'latest' being the most recent release anyway. If there have been special python releases in the past that are now outdated you can simply add a new tag for such images like 'latest-python-outdated'.

luigi311:
I believe the idea for latest for av1an was to pin it to releases only but since av1an was in such rapid development and releases were very far and few in between back then we never actually setup the tagging for latest. I dont think there is a reason we would keep a latest tag for the old python release as so much as changed since then, i dont see why you would utilize that version instead of the new rust versions.

PR #740, opened 2023-03-29
Fix docker example commands

FreezyLemon:
Addresses #733. The av1an:latest image should be updated too, preferably to 0.4.1 (the latest release).

### DIFF
--- a/docs/DOCKER.md
+++ b/docs/DOCKER.md
@@ -14,13 +14,13 @@ The following examples assume the file you want to encode is in your current wor
 Linux
 
 ```bash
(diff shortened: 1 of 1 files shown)


luigi311:
I have no idea how stable the master branch is so i cant comment on if we should default people to use master over latest tags. As for the latest tag i have create a PR #741 so it will update the latest tag whenever a release is created, it wont update it to the 0.4.1 release but it will update it whenever a new release is created. Up to you guys if you want to swap from master to latest for the readme

FreezyLemon:
If both master and latest are recent versions of av1an, the tag used in the example doesn't matter much, I suppose. I just wanted to prevent what happened in the mentioned issue.

I'll close when the #741 gets merged, then.

Issue #719, opened 2023-02-04
target-quality broken


damian101:
Just doesn't do anything. Still works on 0.4.0 branch.

NZDivinity:
Same issue. thought it was user error (first time using av1an) but issue is not present in 0.4.0

BlueSwordM:
Can't reproduce the issue :monkas:

Werve:
After many attempts believing that I got some input args wrong, I confirm that the latest version av1an.exe attached ignores the `--target-vmaf` argument and encodes the video in the normal way e.g. with `--crf`.

You can also check it by passing as argument `--min-q 1 `
the working version 0.4.0 shows the error 
`Error: Condition failed: `min_q > 1` (1 vs 1)` 
while the other one continues normally (ignores the argument)

I believe the versions installed with `cargo` are also affected by the problem

piotrvass:
I can confirm this also happens under new Arch package 0.4.1-1, --target-quality is never run due to changes in commit
 49b3f31 ("Refactor Chunk to be self-contained after generation (#714)", 2023-01-27)

From what I see it will run for some chunk methods, doesn't run 100% for vs pipes (LSMASH).

it's easy to replicate - just run any --target-quality conversion with lsmash. 
Proper log sequence is:
```
DEBUG [av1an_core::broker]  started chunk 00152: 239 frames
DEBUG [av1an_core::broker]  started chunk 00301: 234 frames
DEBUG [av1an_core::broker]  started chunk 00300: 233 frames
DEBUG [av1an_core::broker]  started chunk 00299: 233 frames
DEBUG [av1an_core::target_quality] chunk 00152: P-Rate=2, 239 frames
DEBUG [av1an_core::target_quality] chunk 00152: TQ-Probes: [(90.65, 35), (90.13, 37), (73.09, 60)]
DEBUG [av1an_core::target_quality] chunk 00152: Target Q=37, VMAF=90.00
DEBUG [av1an_core::target_quality] chunk 00300: P-Rate=2, 233 frames
DEBUG [av1an_core::target_quality] chunk 00300: TQ-Probes: [(97.18, 10), (90.06, 34), (89.33, 35)]
DEBUG [av1an_core::target_quality] chunk 00300: Target Q=34, VMAF=90.00
DEBUG [av1an_core::target_quality] chunk 00299: P-Rate=2, 233 frames
[...]

PR #743, opened 2023-03-31
Action: Test target-quality is working correctly.

luigi311:
Addresses #719 

Adds a test to make sure target-quality is actually working by encoding at a target of 95 and 80 and then comparing to make sure the 95 version is bigger than the 80. In the event that it does fail it will also cat the log files to make sure so the developers can see what is happening and adjust the test or fix the issue if there is one. 

In this case we can see that there is no vmaf comparison happening and av1an is just doing a normal chunk encoding thus the two 95 and 80 are exactly the same size and is validated by the logs.


master-of-zen:
That's a good test 👍🏽

master-of-zen:
UNFORTUNATELY, we can't merge it because it doesn't pass tests 🤓

PR #748, opened 2023-04-24
Routine dependency updates

shssoichiro:
This includes a bump for rav1e/av-scenechange, so it may include some scenechange speedups, although I haven't properly benchmarked it.


master-of-zen (approved):
👍🏽

shssoichiro:
Apparently the old rustc doesn't like the new clippy lint being denied...

PR #754, opened 2023-05-09
Move audio size to progress bar OnceCell

master-of-zen:
This removes passing of audio_byte through functions and instead use functions from progress bar module to set and get static audio size bytes


shssoichiro (approved):
Looks a lot better

PR #761, opened 2023-05-26
Split user facing encode args from internals

shssoichiro:
This allows us to set up a constructor for the av1an context and isolate some of the startup checks into av1an's internals.


redzic (approved):

shssoichiro:
Looks like our automerge broke at some point... sad.

PR #764, opened 2023-06-09
Photon noise arguments

superyu1337:
Fixed for newest changes
Old PR description below:
---

Adds photon-noise-width and photon-noise-height, allows for setting the right resolution of the photon noise table in case of resolution adjustments via ffmpeg filters.

Here's a before and after. The source is in 4k, downscaled to 1080p and cropped to 1920x816. 
Both encodes used `--photon-noise 20`

Before:
Before
  
After:
After
[Discord message for more info + testclip](https://discord.com/channels/696849974230515794/702307493623103518/1115606616339910706)


shssoichiro (approved):

shssoichiro:
Can you run `cargo fmt` and then push please? Looks like that's the step that's failing in CI.

superyu1337:
Will do once im home.

Issue #773, opened 2023-07-11
[BISECTED] Accsess violation on vapoursynth input


Quackdoc:
commit 330e6a29396edb6ea0c87c721e43721ed9ada9b3 (bisected and manually tested the previous commit to verify it worked) introduced a regression for one user that causes av1an to be killed by svchost due to accsess violation we narrowed it down to LN 910 `let frame_rate = self.input.frame_rate().unwrap();` by manually adding a before and after check to this. 

av1an isnt providing a backtrace even when compiled we needed to use GFLAGS from windbg since the process was silently exiting. 

it regulairly manages to go through about at least 200 loops before dying. the footage is CFR. tested enforcing it inside of the VS script too with no luck.

still trying to help a user debug it but this is what we have for now, I haven't been able to replicate the issue myself even with the same footage. verified the install environment is the same with VS and same av1an binary.

EDIT: I checked the windows dumps and it looks like this is actually an intermittent vapoursynth crash on the users end. more testing will be done but if nothing else shows up in the next few days ill end up closing this ticket as a VS issue

Quackdoc:
Another user is reporting this crashes during the check framerate step and likewise VS itself is segfaulting on windows

PR #774, opened 2023-07-16
Only get frame rate from Vapoursynth once while creating queue

shssoichiro:
This may mitigate #773, further testing is required to verify. Overall it should slightly improve performance as well.


Quackdoc:
one user is reporting this to fix their issue.

redzic (approved):

Issue #745, opened 2023-04-16
Chunk methods introduce image glitches


mxsrm:
I've been testing with multiple sources and all the chunk methods and **only** `select` is not creating this problem. lsmash, ffms2 and hybrid do.

If av1an selects chunks to close to a keyframe, sometimes vaporsynth will create blocky artifcats:

```
vspipe -s 1971 -e 1982 loadscript.vpy - -c y4m | mpv -
```
![mpv0002-good_lsmash y4m](https://user-images.githubusercontent.com/7951225/232312360-ed80082d-9468-4f95-842a-09cb1abf2d15.jpg)
(like in the source material)

```
vspipe -s 1972 -e 1982 loadscript.vpy - -c y4m | mpv -
```
![mpv0001-bad_lsmash y4m](https://user-images.githubusercontent.com/7951225/232312378-12a925f1-ee7a-4be0-a416-7d88ee7bdb78.jpg)

As said, this happens over multiple source materials from different origins. If you encode the movie en-block (e.g. x264, x265, svt-av1) no glitches appear. Only happens in chunked encoding and using anything but select, which is of course insanely slow.

Anybody have a workaround?

Version information:
```
> av1an --version
av1an 0.4.1-unstable (rev 45a927d) (release)

* Compiler
  rustc 1.68.0 (LLVM 15.0)

* Target Triple
  x86_64-unknown-linux-gnu

* Date Info
  Commit Date:  2023-04-06

* VapourSynth Plugins
  systems.innocent.lsmas : Found
  com.vapoursynth.ffms2  : Found

> vspipe --version
VapourSynth Video Processing Library
Copyright (c) 2012-2022 Fredrik Mellbin
Core R61
API R4.0
API R3.6
Options: -
```

mxsrm:
Closed since no one cares that this completely breaks av1an in 99% of the use cases.

mxsrm:
BluRay remuxes in mkv format, either H264 (for 1080p) or H265 (for 2160p). I saw this bug over multiple different sources.

Simulping:
the only way to prevent this is to re encode with x264 qp 0 as a lossless intermediary. although this can be annoying for UHD Blu-rays since H.264 does not have any HDR specifications

shssoichiro:
You don't have to worry about HDR specifications in the lossless intermediary, as long as they get copied over to the final encode one way or another.

That being said it would be nice if there was a way to fix this issue... though I don't currently know what that fix would look like.

[...]

PR #776, opened 2023-07-16
Add DGDecNV and BestSource chunk method

Simulping:
As the title suggests, adds two new chunking methods to Av1an.

The docs is a bit general and could be worked on a bit since I don't really want to say anything that could be misinformation about the technical details. The current chunk method order now is: DGDecNV > lsmash > ffms2 > bestsource > hybrid

[DGDecNV](https://www.rationalqm.us/dgdecnv/dgdecnv.html) requires an external program called "dgindexnv" to index the video first, though its [already bundled](https://www.rationalqm.us/dgdecnv/binaries/) with the binaries and plugins itself provided, and as the name suggests, it requires an NVIDIA GPU. Which also requires PureVideo, but that's ancient technology so everyone should have it anyway. And only decodes AVC/HEVC/MPG/VC-1.

BestSource is similar to select, its slow but should decode everything

 
and from my observation and testing, both methods successfully "resolve" #745 
here's a test clip where gray frames keep appearing: [https://cdn.discordapp.com/attachments/587033245061873759/1126937326333997076/CutJJBA1-7.mkv](https://cdn.discordapp.com/attachments/587033245061873759/1126937326333997076/CutJJBA1-7.mkv)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,9 @@ Prerequisites:
 Optional:
 
 - [L-SMASH](https://github.com/AkarinVS/L-SMASH-Works) VapourSynth plugin for better chunking (recommended)
+- [DGDecNV](https://www.rationalqm.us/dgdecnv/dgdecnv.html) Vapoursynth plugin for very fast and accurate chunking, `dgindexnv` executable needs to be present in system path and an NVIDIA GPU with CUVID 
 - [ffms2](https://github.com/FFMS/ffms2) VapourSynth plugin for better chunking
-- [DGDecNV](https://www.rationalqm.us/dgdecnv/dgdecnv.html) Vapoursynth plugin, `dgindexnv` executable present in system path, and an NVIDIA GPU for very fast chunking
+- [bestsource](https://github.com/vapoursynth/bestsource) Vapoursynth plugin for slow but accurate chunking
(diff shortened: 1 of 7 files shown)


Quackdoc:
hwdec on a lot of devices is for sure not accurate a lot of times, that being said, having a super slim colorshift is probably a lot better then massive gray squares. IMO any hwdec should never be considered "accurate"

shssoichiro:
I don't love having that as the default... but I'm not sure if there's a better option since lsmash kind of fails on VC1 and certain other codecs, and bestsource is really, really slow... I may brainstorm or google a bit when I have time.

Bl4cKn1gh7:
I also don't think that having DGDecNV as the default makes that much sense, as it's not like everyone has an Nvidia GPU on their system. That being said, I think it doesn't hurt to add it as long as any concerns are included in the help page of Av1an, e.g. not mathematically lossless, requires nvidia purevideo, etc.

I think it would be a good addition to help solve the grey frame problem, and if someone has hardware that they want to use, as long as they're aware of what the nuances are, they should be able to use it. As long as it doesn't create problems for the rest of the program of course.

Simulping:
@shssoichiro @Bl4cKn1gh7 I'll restore the original order and edit the PR then, later

shssoichiro (approved):

master-of-zen:
we have a certified tabulation moment

PR #795, opened 2023-12-24
unimportant improvements

damian101:
The removed parameters all don't do anything, even at cpu-used 4, and deltaq-mode 0 doesn't do anything if TPL model is disabled anyway.
Max-partition-size=32 just causes a slowdown in modern aomenc. Disabling additional keyframe placement is generally faster with aomenc and very suited for chunked encoding.


master-of-zen:
@shssoichiro your opinion on the matter?

shssoichiro:
This seems fine for the target quality command, it's unlikely to make a significant difference in quality from that standpoint.

shssoichiro (approved):

master-of-zen:
LGTM

Issue #662, opened 2022-09-06
Does `--video-params` disable default parameters and prevent target quality?


mqudsi:
I observed that if I run `av1an -e svt-av1 --target-quality=xx ....` the output includes 

> Params: --preset 4 --keyint 240 --rc 0 --crf 25

But if I run `av1an -e svt-av1 -v '--preset 10' --target-quality=xx ...`, the output just says

> Params: --preset 10

To me, this implies that the presence of an explicitly specified `--video-params` is preventing `av1an` from supplying its own parameters. My reading of the help output led me to assume that `--video-params` would be _in addition to_ to and not _instead of_ the default per-encoder video parameters av1an used to get its job done.

IMHO, av1an should always pass its video parameters unless the `-v` includes the same parameter (in which case the manually supplied parameter should win) - this would be for things like `--preset`, `--keyint`, etc. But if `-v` included `--crf` and `--target-quality` was used, I would expect av1an to throw an error complaining that `--crf` and `--target-quality` cannot be used together.

clidx:
I agree with both of your points, but would just add that `--crf` and `--target-quality` not conflicting is a holdover from before when `--crf` had to be manually specified in the params so that av1an could change it. Now, not having them conflict adds unnecessary ambiguity about what av1an is actually doing.

FreezyLemon:
When av1an shows you

> Params: --preset 4 --keyint 240 --rc 0 --crf 25

it just means that those are the "base" parameters it uses for encoding. Target quality mode decides on a CRF value for every chunk based on your target quality setting, which means your actual encoder parameters will be `--preset 4 --keyint 240 --rc 0 --crf x`, with x possibly being different for every chunk.

If you use custom `--video-params`, that just means that your "base" parameters are different. It does *not* mean that av1an won't be able to change the CRF value. If you use `--video-params "--preset 8 --keyint 0 --lookahead 60"`, then av1an will automatically add the required stuff to make it work: `--preset 8 --keyint 0 --lookahead 60 --crf x`. Again, x being decided by the target quality algorithm.

mqudsi:
Thanks for the explanation! I definitely wasn’t able to intuit  that from the output.

PR #798, opened 2024-01-05
Improve setting default encode parameters

FlyingWombat:
This improves upon setting default encoding parameters. The gist of this pull request is improving the tile and worker estimation, and keeping those values when setting video-params. My main use-case for this is that I sometimes have batches of mixed resolution videos (e.g. 1080p and 4k) that I want to encode with the same settings -- and not have to go through one by one to specify tile count.

Some of the commits are loosely related, and I wasn't sure if I should make separate pull requests for them, so they are included here. I am a beginner with rust, so please correct me if the formatting is bad, or if there is a better way to implement the changes.

### DIFF
--- a/av1an-core/src/context.rs
+++ b/av1an-core/src/context.rs
@@ -4,7 +4,7 @@ use std::collections::BTreeSet;
 use std::convert::TryInto;
 use std::ffi::OsString;
 use std::fs::File;
-use std::io::Write;
+use std::io::{IsTerminal, Write};
 use std::path::{Path, PathBuf};
 use std::process::{exit, Command, Stdio};
 use std::sync::atomic::{self, AtomicBool, AtomicUsize};
@@ -12,6 +12,7 @@ use std::sync::{mpsc, Arc};
 use std::thread::available_parallelism;
 use std::{cmp, fs, iter, thread};
 
+use ansi_term::{Color, Style};
 use anyhow::{bail, Context};
 use av1_grain::TransferFunction;
 use crossbeam_utils;
@@ -263,14 +264,36 @@ impl Av1anContext {
       }
       self.args.workers = cmp::min(self.args.workers, chunk_queue.len());
 
-      info!(
-        "Queue {} Workers {} Passes {} Encoder {}\nParams: {}",
-        chunk_queue.len(),
-        self.args.workers,
-        self.args.passes,
-        self.args.encoder,
-        self.args.video_params.join(" ")
-      );
+      if std::io::stderr().is_terminal() {
+        eprintln!(
+          "{}{} {} {}{} {} {}{} {} {}{} {}\n{}: {}",
+          Color::Green.bold().paint("Q"),
+          Color::Green.paint("ueue"),
+          Color::Green.bold().paint(format!("{}", chunk_queue.len())),
+          Color::Blue.bold().paint("W"),
+          Color::Blue.paint("orkers"),
+          Color::Blue.bold().paint(format!("{}", self.args.workers)),
+          Color::Purple.bold().paint("E"),
+          Color::Purple.paint("ncoder"),
+          Color::Purple.bold().paint(format!("{}", self.args.encoder)),
+          Color::Purple.bold().paint("P"),
+          Color::Purple.paint("asses"),
+          Color::Purple.bold().paint(format!("{}", self.args.passes)),
+          Style::default().bold().paint("Params"),
+          Style::default()
+            .dimmed()
+            .paint(self.args.video_params.join(" "))
+        );
+      } else {
+        eprintln!(
+          "Queue {} Workers {} Encoder {} Passes {}\nParams: {}",
+          chunk_queue.len(),
+          self.args.workers,
+          self.args.encoder,
+          self.args.passes,
+          self.args.video_params.join(" ")
+        );
+      }
 
       if self.args.verbosity == Verbosity::Normal {
         init_progress_bar(self.frames as u64, initial_frames as u64);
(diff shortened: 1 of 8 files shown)


shssoichiro:
Thanks, I need to look through more in detail but a lot of this looks like an improvement.

damian101:
"determine_workers: consider source resolution and tiles" 
I don't want Av1an to change encoder defaults regarding tiling, if that is what's done here, simply because I consider it unexpected behavior. At least not without some message that it did add those parameters, overriding encoder defaults.

Setting keyframe interval to unlimited by default, however, sounds much more like a good idea, as users will expect Av1an to mess with keyframe placement anyway, and you usually want to control the max keyframe distance through Av1an chunking. Less sure about disabling automatic keyframe placement altogether by default. The thing is, I have used Av1an for single-chunk encoding in the past, simply expecting encoder defaults to be used...

I would be completely fine with any reasonable encoder parameters being added by default, if a message is always printed listing those parameters, so users are aware of them. I think that would be a good idea even without these additional changes to Av1an's default encoder parameters.

FlyingWombat:
The default tiling estimation was already there, this pull request doesn't change that -- just makes it actually useful.  Av1an already, by default will determine tiles and set encoder parameters accordingly. But as is, if --video-params is specified, all default parameters set by Av1an are ignored, such as the (useful) tile estimation. What this pull request fixes is that now the tiling estimation can be actually used with user-specified encoder parameters. The "settings: merge video_params..." commit is what does this, and also has the --force flag ignore Av1an's default encoder parameters altogether.  If we don't want Av1an to set encoder parameters automatically, then we should remove the tiling estimation and default parameters code altogether.

Considering tiles in worker estimation is better because (except for svt-av1) tiling changes the CPU resources used by each worker. Not considering tiles in worker estimation would over-commit the CPU and use more RAM than necessary.

Av1an already prints the encoder parameters being used, and I added a message for that in the log file in this pull request.

damian101:
More tiles reduce efficiency. Aomenc by default only uses as many as necessary. Tiling is something I definitely want to tune myself, not have Av1an set them automatically if I don't clutter my encoding parameters with `--tile-rows=0 --tile-columns=0` to protect myself from it.
Also, as far as I understand the code, with aomenc this would result in:
1080p: 1*2=2 tiles
1440p: 1*2=2 tiles
2160p: 2*4=8 tiles
But it's not common at all to do it this way. 2 tiles for 2160p appears to be the common way to do tiling, which would require the 720 in the code to be replaced with maybe 1280, unless I got the math wrong...

FlyingWombat:
You can still have the old behavior, where Av1an doesn't set default parameters at all. This pull request has the --force flag ignore all of Av1an's default encoder parameters. Do you think I should put this option in a separate flag, in case a user wants to ignore default arguments, but still have the syntax check?

We could add a flag to opt-out of just tile estimation e.g. `--no-tile`, if you think a significant portion of people would use it.

The main reason I made the change to merge video_params into defaults, is because I wanted to use the tiling and worker estimation. Tiling and worker estimation, and all the default parameter code for that matter, is unused when --video-params is set -- which pretty much everyone does to at least set --crf. There has been confusion about this behavior before, see #662 . This pull request fixes that. Without default parameter merging, I'd have to clutter my encoding parameters with e.g. `--keyint 0 --scd 0 --rc 0`, which only need to be set because I'm using av1an, and I'd have to make an external wrapper for av1an to estimate tiles based on resolution.

We may need to change the numbers used for tile estimation. I also thought it was too many tiles, and changed it in my local copy of Av1an.

damian101:
My main issue is the encoder doing things people do not expect it to do, which here mostly concerns tiling.
I also don't think threading is a significant issue for most people with Av1an chunked encoding. The main reason people use two tiles for 4K appears to be that it makes decoding easier.
And, if there was a message listing all automatically added parameters, I wouldn't mind any reasonable parameters being added by default...

FlyingWombat:
@damian101  Do you have any specific changes to propose for this pull request? I'm not sure what to do with your feedback.

Does the added behavior for the `--force` flag meet your needs?
Or, would adding a new flag specifically to control tile estimation be better? (but then we're adding another cli flag) E.g. `--no-tile` to disable tile estimation; or `--auto-tile` to enable tile estimation. Should tile estimation be enabled by default, or disabled by default?

The encoding parameters are printed to the screen and the log file.

damian101:
> @damian101  Do you have any specific changes to propose for this pull request? I'm not sure what to do with your feedback.
> 
> Does the added behavior for the `--force` flag meet your needs?
> Or, would adding a new flag specifically to control tile estimation be better? (but then we're adding another cli flag) E.g. `--no-tile` to disable tile estimation; or `--auto-tile` to enable tile estimation. Should tile estimation be enabled by default, or disabled by default?
> 
> The encoding parameters are printed to the screen and the log file.

My strong preference would be to not do additional tiling by default, but have a tiling parameter for those who like this feature. Especially considering there is no clear data-backed consensus in the community on how additional tiling should ideally be done in the first place, if at all, last time people did some testing was ages ago on now very outdated aomenc...

Regarding all other aspects of this PR, I at the least have no strong objections.

damian101:
I might be wrong, but I don't see log2 applied to the return of function `calculate_tiles`, which to my understanding calculates a raw number of tiles. And unlike rav1e, aomenc and svt-av1 take the log2 of the actual number of tiles.

FlyingWombat:
@damian101  It took me a while to get back to this, but per your request, I've added a "--tile-auto" flag. Av1an will now only do tile estimation (and automatic setting of tiling encode params) if this flag is present.

log2 conversion is done in encoder.rs::get_default_arguments for applicable encoders.

shssoichiro (approved):
Sorry for the very long delay on reviewing this, since CI was broken for quite a long time. I made a few changes, notably reverting the removal of the terminal colorization, and splitting `no-defaults` into a separate flag from `force`, since I don't think it makes sense to have `force` do both of those things.

PR #803, opened 2024-01-08
Mention silent truncation of probing_rate>4

t-nil:
In target_quality.rs, `fn adapt_probing_rate()` is used to truncate the probing rate. This is not mentioned in the parameter description.

### DIFF
--- a/av1an/src/main.rs
+++ b/av1an/src/main.rs
@@ -620,9 +620,14 @@ pub struct CliOpts {
     #[clap(long, default_value_t = 4, help_heading = "Target Quality")]
     pub probes: u32,
 
-    /// Framerate for probes, 1 - original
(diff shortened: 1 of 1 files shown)


FreezyLemon:
This is surprising behaviour, and I did not know it worked like that. It might be even better to max it out at 4 and give a runtime warning..

shssoichiro:
I second the above suggestion, I'd rather have less surprising behavior for the end user if possible. If it's possible to update this PR to make that change, it would be appreciated.

t-nil:
```bash
> ~/.cache/cargo_target/debug/av1an --probing-rate 5
error: invalid value '5' for '--probing-rate <PROBING_RATE>': 5 is not in 1..=4

For more information, try '--help'.
```

Should be mergable now

shssoichiro (approved):

PR #804, opened 2024-01-14
Update dependencies

shssoichiro:
Apparently it's been a while since we did this. Pulls in rav1e 0.7 among other items.


shssoichiro:
The latest Ubuntu image ships with Rust 1.66... what a terrible packaging system. I'll have to rework the CI to do something about this. :facepalm:

redzic (approved):

PR #814, opened 2024-02-25
Implement alternate input for scene detection V2

BlueSwordM:
Context: https://github.com/master-of-zen/Av1an/pull/631

This is the refactored version of the patch.
I've build it and it seems to be fine.

Just need to do some test encodes.

### DIFF
--- a/av1an/src/main.rs
+++ b/av1an/src/main.rs
@@ -121,6 +121,10 @@ pub struct CliOpts {
   #[clap(short, required = true)]
   pub input: Vec<PathBuf>,
 
+  /// Alternate input for scene detection (must have the same frame count)
+  #[clap(short = 'I')]
+  pub input_sc: Option<PathBuf>,
+
   /// Video output file
   #[clap(short)]
   pub output_file: Option<PathBuf>,
@@ -634,6 +638,7 @@ pub fn parse_cli(args: CliOpts) -> anyhow::Result<Vec<EncodeArgs>> {
     };
 
     let input = Input::from(input);
+    let input_sc = args.input_sc.as_ref().map(Input::from);
 
     let video_params = if let Some(args) = args.video_params.as_ref() {
       shlex::split(args).ok_or_else(|| anyhow!("Failed to split video encoder arguments"))?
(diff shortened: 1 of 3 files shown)


av1an/src/main.rs
```diff
@@ -121,6 +121,10 @@ pub struct CliOpts {
   #[clap(short, required = true)]
   pub input: Vec<PathBuf>,
 
+  /// Alternate input for scene detection (must have the same frame count)
```
FreezyLemon, line 124: What happens when the two inputs don't have the same frame count?

lastrosade: > What happens when the two inputs don't have the same frame count?

This needs a check, With at least a warning to prevent future hair pulling.

BlueSwordM: This is indeed something that should be added to prevent... annoying reports in the future.

BoatsMcGee:
Howdy. I've been working on an overhaul of Target Quality and have come across a similar desire for an alternative input that is simpler to compute for transient tasks like Scene Detection and Target Quality. Users could have very compute-heavy VapourSynth scripts as the input and would currently need to run Scene Detection separately to generate the Scenes JSON file. However, this is currently not possible for Target Quality as each chunk is processed and encoded in the same step.

I'd like to implement this change a bit differently. The alternative input should be renamed to `proxy` (`-p`, `--proxy`) - just like how video editing programs create simpler or smaller proxies of the larger original input. This is more generic and a familiar term for people who use adjacent software. A user could provide a complicated script as the input and the original video itself as the proxy for use in Scene Detection and Target Quality. Of course, the input and proxy must be checked to ensure they have the same frame count to avoid potential mismatch errors. That can be done in or perhaps after `parse_cli()` or we may want to implement a dedicated validation step. I would prefer the latter for reasons mentioned below.

Let me know what you think of this suggestion. Fair warning: this is low priority for me as I'm busy with Target Quality (I've gotten ssimulacra2 and butteraugli both working and have plans for more metrics and will draft a PR for it eventually) and potentially many other supporting changes.

Thanks,
\- Boats

shssoichiro:
Closing in favor of #1028

PR #817, opened 2024-03-24
append to log file

jcj83429:
When resuming, append to the log file instead of clearing its old contents.

I tested with and without resume.


shssoichiro (approved):
Looks good, thanks. The CI failure seems unrelated, probably some new clippy lint. I'll get that fixed but you'll need to rebase after.

shssoichiro:
Blocked by #819

PR #823, opened 2024-04-23
Fix vmaf flag check with target-quality

### LINKED ISSUE #790: Using --target-quality forces --vmaf
Seems like whenever you use target-quality it forces the vmaf that happens after encoding finishes. You can see the : : VMAF Run line at end after the encode has finished when i never added in the --vmaf flag.

luigi311:
Fixes issue where vmaf plot was always ran if target-quality was set, this does not allow you to run vmaf plot without target-quality as it seems like vmaf plot was designed around target-quality at some point so --vmaf without --target-quality will do nothing. Decoupling the two seems like it will be a lot of work so I am adding this for now as it atleast closes #790 and the decoupling can happen later

### DIFF
--- a/av1an-core/src/context.rs
+++ b/av1an-core/src/context.rs
@@ -360,17 +360,19 @@ impl Av1anContext {
           temp_res = tq.vmaf_res.to_string();
         }
 
-        if let Err(e) = vmaf::plot(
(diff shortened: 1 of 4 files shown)


master-of-zen:
👍🏽

master-of-zen (approved):

PR #825, opened 2024-04-23
Add support for vmaf-subsample

### LINKED ISSUE #806: [Feature] Pass extra VMAF args (like n_subsample, pool)
[ab-av1](https://github.com/alexheretic/ab-av1) has a switch to pass extra ffmpeg libvmaf parameters. I would really love to try out `n_subsample`, as `--probing-rate` gives near unusable results, but the VMAF switch could atleast speed up VMAF calculation at faster presets, where it matters.

luigi311:
Add vmaf-subsample as an alternative to probing-rate as it is more accurate while still providing a speed benefit compared to without it 
Closes #806

Looks like probe-rate 3 in this example is useless as it messes up the quality so bad it just uses 8 crf which was my min while subsample doesnt seem to run into any issues with setting it a bit to high.

Testing on a 30 second 1080p clip with little motion
Target quality of 96 using x265 fast with extra-split 240 comes out to 7 scenes

# default
real    3m34.966s
user    24m37.871s
sys     2m58.700s

default

# probe 2
real    1m45.946s
user    11m48.243s
sys     1m41.285s

probing-rate_2

# probe 3
real    1m44.766s
user    10m12.201s
sys     1m30.548s

probing-rate_3

# subsample 3
real    2m35.967s
user    17m47.618s
sys     2m25.497s

vmaf-subsample_3

# subsample 9
real    2m8.224s
user    14m7.415s
sys     1m50.762s

vmaf-subsample_9

### DIFF
--- a/av1an-core/src/context.rs
+++ b/av1an-core/src/context.rs
@@ -368,6 +368,7 @@ impl Av1anContext {
             temp_res.as_str(),
             tq.vmaf_scaler.as_str(),
             1,
+            1,
             tq.vmaf_filter.as_deref(),
             tq.vmaf_threads,
           ) {
--- a/av1an-core/src/target_quality.rs
+++ b/av1an-core/src/target_quality.rs
@@ -25,6 +25,7 @@ pub struct TargetQuality {
   pub vmaf_threads: usize,
   pub model: Option<PathBuf>,
   pub probing_rate: usize,
+  pub probing_subsample: usize,
   pub probes: u32,
   pub target: f64,
   pub min_q: u32,
@@ -252,6 +253,7 @@ impl TargetQuality {
       &self.vmaf_res,
       &self.vmaf_scaler,
       self.probing_rate,
+      self.probing_subsample,
       self.vmaf_filter.as_deref(),
       self.vmaf_threads,
     )?;
--- a/av1an-core/src/vmaf.rs
+++ b/av1an-core/src/vmaf.rs
@@ -122,6 +122,7 @@ pub fn plot(
   res: &str,
   scaler: &str,
   sample_rate: usize,
+  subsample: usize,
   filter: Option<&str>,
   threads: usize,
 ) -> Result<(), Box<EncoderCrash>> {
@@ -156,6 +157,7 @@ pub fn plot(
     res,
     scaler,
     sample_rate,
+    subsample,
     filter,
     threads,
   )?;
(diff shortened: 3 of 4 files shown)


luigi311:
I just tested with a heavy motion action scene and while it seems like they all chose the min crf of 8 the logs show a completely different picture. Looks like subsample is actually smart enough to limit itself so it doesnt try to go to high during high movement scenarios where as probe doesnt and probe 3 shows a min vmaf of 0.

Testing on a 30 second 1080p clip with lots of motion
Target quality of 96 using x265 fast with extra-split 240 comes out to 10 scenes

# default
real    3m1.954s
user    25m15.300s
sys     1m31.100s

Min VMAF: 91.71 

# probe 2
real    2m39.683s
user    20m36.260s
sys     1m27.650s

Min VMAF: 91.97

# probe 3
real    2m19.869s
user    17m33.692s
sys     1m18.733s

Min VMAF: 0

# subsample 3
real    3m0.207s
user    23m29.094s
sys     1m36.000s

Min VMAF: 91.54

# subsample 9
real    3m0.877s
user    23m13.567s
sys     1m33.432s

Min VMAF: 91.54

luigi311:
renamed to probing-subsample as it only applies to probes and not the final vmaf similar to probing-rate

luigi311:
this also only provides a benefit if vmaf is the slowest part of the target quality process so you will not see many changes using --probe-slow if you are using a slow preset.

shssoichiro:
I think the code looks fine, but I don't normally use the VMAF features to know more about how useful this is. Thoughts @redzic @master-of-zen ?

damster101:
> I just tested with a heavy motion action scene and while it seems like they all chose the min crf of 8 the logs show a completely different picture. Looks like subsample is actually smart enough to limit itself so it doesnt try to go to high during high movement scenarios where as probe doesnt and probe 3 shows a min vmaf of 0.

This is very much expected, because the difference between probing-rate and the vmaf subsample feature is simply that with the latter, the motion component still runs on all frames, avoiding distorted offsets for the motion component.
If this becomes an option, the probing-rate feature will cease to be useful for anything but high framerate sources, in that case to increase vmaf motion accuracy, as vmaf is framerate-agnostic.

master-of-zen:
> I think the code looks fine, but I don't normally use the VMAF features to know more about how useful this is. Thoughts @redzic @master-of-zen ?

Good

PR #835, opened 2024-05-06
Bump MSRV to 1.70

shssoichiro:
This is already the actual MSRV, but it was not updated in the Cargo config files, so this change is needed to fix CI.


master-of-zen (approved):
👍🏽

Issue #837, opened 2024-05-10
SVT-AV1-PSY --target-quality support?


NZDivinity:
Currently the --target-quality parameter for Av1an does not work with the PSY fork of SVT-AV1 - errors out.
--target-quality does work when it is used in conjunction with --probe-slow.

The same Av1an command (without probe slow) works fine with regular SVT-AV1.

Any chance this parameter could be supported by SVT-AV1-PSY?

Edit : My bad, was a bit quick posting this issue. see Trix's reply below with context on possible cause.

trixoniisama:
You could have given some additional context since we've started to troubleshoot it on the AV1 discord server...

It is likely that SVT-AV1-PSY is not recognized as modern SVT-AV1 and as such outdated parameters are fed to the encoder which leads to crashes.

The usefulness of keeping compatibility with incredibly old SVT-AV1 versions isn't evident to me. I think that to avoid further issues like this in the future, it would be better to remove support for these ancient versions. It will avoid false positives with potential forks and newer version numbers.

gitoss:
> Currently the --target-quality parameter for Av1an does not work with the PSY fork of SVT-AV1 - errors out.

SVT-AV1-PSY has been updated to current SVT-AV1 and is working just fine, so this issue can be closed.

NZDivinity:
Unfortunately this issue is back as of SVT-AV1-PSY's recent (pre-release) update v2.2.0-rc1

Plamper:
This issue seems to stem from the set version tag in [SVT-AV1-PSY's Cmakelists.txt](https://github.com/gianni-rosato/svt-av1-psy/blob/v2.2.1/CMakeLists.txt). If someone runs into this issue while self compiling, just change the line `set(VERSION_TAG "rc2")` to `set(VERSION_TAG "")`. The release binaries seem to have already changed this.

PR #838, opened 2024-05-10
Fix new svt-av1 versions 2.x.x being treated as old svt-av1

damian101:
Invalid use of `1..` in a match function, as it has no upper bound. Replaced with a wildcard pattern to simply catch all remaining patterns, which I think should work fine here.


shssoichiro:
This is a little confusing. Matches are supposed to be exhaustive, so `1..` should be catching everything 1 and greater. What bug is occurring here? This should be a compile-time error if the match is not exhaustive.

trixoniisama:
#837

shssoichiro:
I see, however, I don't understand why this fixes the issue. As I mentioned, `1..` and `_` should be equivalent here. If they're not, then that's a very unexpected behavior because the compiler should be throwing errors if we currently have a non-exhaustive match statement.

PR #846, opened 2024-05-22
Updates to documentation

0xBA5E64:
This is a collection of work-in-progress updates to the Av1an docs, spanning the README & Book.

## Corrections / comments are greatly appreciated!


master-of-zen:
@shssoichiro Looks good to me, your opinion?

shssoichiro:
I'm good with it also :+1:

master-of-zen:
One nit pick, i liked previous big screenshot)

0xBA5E64:
> One nit pick, i liked previous big screenshot)

Fair enough @master-of-zen , I also did like seeing it, both for the flex and since it shows off Av1an's quite striking progress-meter. However, felt that the project description should be right at the top so that you immediately get to know what the project is about, but this could of course also be remedied by simply moving the screenshot to below the first paragraph too, i just happened to like the more compact "complex" look.

0xBA5E64:
Other than that, I still want to re-write the CLI Reference in the book to be a bit more, readable, before I ready this up for review.

I've been experimenting with some different ways of doing that, most notably by turning those pages into HTML Tables but, it makes the docs a nightmare to maintain, and I fear it'd desentivize people from keeping it in sync with the in-code helptext.

Perhaps the "CLI-Reference" should literally be auto-generated on the docs compiling to be the output of `av1an --help`, on a single page, just so that users have something to easily `Ctrl`+`F` though if they want, with additional written guides to the different parts of av1an's CLI standalone in the docs...

Vernoxvernax:
Aomenc:
In the description of `--lag-in-frames` it says `35` is the highest value and `19` the default, but in the examples it reads `up to a limit of 48 (default is 35)`.
Don't know which one is correct.

Issue #850, opened 2024-06-08
Dockerfile: error: target not found: vapoursynth-plugin-lsmashsource


Extarys:
It appears that `vapoursynth-plugin-lsmashsource ` was removed from arch repo:
vapoursynth-plugin-lsmashsource 1:A.5b-2 is no longer available

PR #851, opened 2024-06-08
Remove vapoursynth-plugin-lsmashsource

Extarys:
Plugin doesn't exist on Arch repo anymore.

Fix https://github.com/master-of-zen/Av1an/issues/850

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM archlinux:base-devel AS base
 RUN pacman -Syu --noconfirm
 
 # Install dependancies needed by all steps including runtime step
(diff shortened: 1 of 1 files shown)


shssoichiro:
I wouldn't necessarily recommend bestsource as the first choice. Its indexing is very slow, and it tends to sometimes do weird things with guessing framerate. Though it would probably be a good idea to have instructions for each of the possible source plugins.

Which, for lsmashsource would be to get the git package from the AUR. It's strange to me that the Arch maintainers removed the stable version from the official repos.

shssoichiro:
It looks like we need this removed in order to fix the CI build... We'll need to figure out an alternative solution at some point. I think the idea is setting up the build to get `vapoursynth-plugin-lsmashsource-git` from the AUR, but that can come in a later change. For now, I'll go ahead and merge this so the CI failure isn't blocking everything.

shssoichiro (approved):

PR #853, opened 2024-06-10
Use updated av-scenechange with threaded ffmpeg decoder

### LINKED ISSUE #852: Scene detection performance regression
Scene detection speed has a performance regression because of commit https://github.com/master-of-zen/Av1an/commit/5184c0504e7d7af9a2a564558d9bf0c9bfd0474d
This happens with any video (that I've ever tested at least)
This issue is especially worse on high bitrate video
How to reproduce:
1. Run a test encode on a video with av1an commit https://github.com/master-of-zen/Av1an/commit/5184c0504e7d7af9a2a564558d9bf0c9bfd0474d, the encoder or the encoder parameters doesn't matter because we're only comparing the scene detection speed. Here is an example command that I used, `av1an -i input.mkv -o output.mkv -e svt-av1`, you can cancel the encode afterward if the scene detection is done
2. Run a test encode on a video with av1an prior to commit https://github.com/master-of-zen/Av1an/commit/5184c0504e7d7af9a2a564558d9bf0c9bfd0474d. In this case I used av1an commit https://github.com/master-of-zen/Av1an/commit/a506a924f2c98dca7512eb0ba8daf0e68c51b42a
3. Compare the speed of the scene detection, and you can see that av1an commit https://github.com/master-of-zen/Av1an/commit/5184c0504e7d7af9a2a564558d9bf0c9bfd0474d is slower at scene detection

shssoichiro:
Fixes #852

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,9 +213,9 @@ dependencies = [
 
 [[package]]
 name = "av-scenechange"
-version = "0.12.0"
+version = "0.12.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a5fa17f98b2134d5459bbf2e69b297b1be2047a3f617af34fad7ecca362c6cb5"
+checksum = "f0707e4b0e933849f2955134399e268cbefb96fb9524b0aa3fb612c3d3ddacf5"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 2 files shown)


redzic (approved):
Thanks!

PR #855, opened 2024-06-21
Update compile docs

Uranite:
Updating it since its out of date
Python is not needed if vapoursynth is portable


shssoichiro (approved):
Thanks!

PR #861, opened 2024-07-23
Use absolute instead of canonicalize for better Windows path handling

### LINKED ISSUE #860: Change path handling in vapoursynth.rs for compatibility
Currently the source file for vapoursynth is being canonicalized immediately (`std::fs::canonicalize`) which creates extended (UNC-like) paths on Windows. Depending on the compiler and functions used, code written in other languages may fail to understand these.

Changing the line `let source = std::fs::canonicalize(source)` to use [`std::path::absolute`](https://doc.rust-lang.org/std/path/fn.absolute.html#windows-paths) instead would fix and prevent future occurrances of e.g. vapoursynth/bestsource#59 altogether.

```rs
let source = if cfg!(target_os = "windows") {
    (std::path::absolute(source) 
} else {
    std::fs::canonicalize(source) 
};
```

shssoichiro:
Closes #860

### DIFF
--- a/av1an-core/Cargo.toml
+++ b/av1an-core/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "av1an-core"
 version = "0.4.1"
-rust-version = "1.70"
+rust-version = "1.79"
 edition = "2021"
 authors = ["Zen <master_of_zen@protonmail.com>"]
 description = """
--- a/av1an-core/src/util.rs
+++ b/av1an-core/src/util.rs
@@ -1,4 +1,7 @@
-use std::path::{Path, PathBuf};
+use std::{
+  io,
+  path::{absolute, Path, PathBuf},
+};
 
 /// Count the number of elements passed to this macro.
 ///
@@ -131,6 +134,15 @@ pub fn read_in_dir(path: &Path) -> anyhow::Result<impl Iterator<Item = PathBuf>>
       }
     })
   }))
+}
+
+#[inline]
+pub(crate) fn to_absolute_path(path: &Path) -> io::Result<PathBuf> {
+  if cfg!(target_os = "windows") {
+    absolute(path)
+  } else {
+    path.canonicalize()
+  }
 }
 
 #[cfg(test)]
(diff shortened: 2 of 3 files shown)


Uranite:
It seems that the cachepath is still canonicalized
notepad++_960_2024-07-28-1722163870
as a result, when using bestsource, the path for the cache is wrong (see how the path repeats)
`C:\Users\redacted\Videos\osu!\osu! 2024.05.25 - 12.33.34.01\split\cache.json\Users\redacted\Videos\osu!\osu! 2024.05.25 - 12.33.34.01.mp4.0.bsindex`

Uranite:
Anyway, here are my proposed change for this PR
By using `to_absolute_path` for `cache_file` we can safely remove `use path_abs::PathAbs;`
Also, bestsource has stopped generating json indexes since a long time now, so I changed it to bsindex
```diff
diff --git a/av1an-core/src/vapoursynth.rs b/av1an-core/src/vapoursynth.rs
index 7c8484e..4a5f1f5 100644
--- a/av1an-core/src/vapoursynth.rs
+++ b/av1an-core/src/vapoursynth.rs
@@ -5,7 +5,6 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, bail};
 use once_cell::sync::Lazy;
-use path_abs::PathAbs;
 use std::process::Command;
 use vapoursynth::prelude::*;
 use vapoursynth::video_info::VideoInfo;
@@ -201,13 +200,13 @@ pub fn create_vs_file(
 
   let mut load_script = File::create(&load_script_path)?;
 
-  let cache_file = PathAbs::new(temp.join("split").join(format!(
+  let cache_file = to_absolute_path(&temp.join("split").join(format!(
     "cache.{}",
     match chunk_method {
       ChunkMethod::FFMS2 => "ffindex",
       ChunkMethod::LSMASH => "lwi",
       ChunkMethod::DGDECNV => "dgi",
-      ChunkMethod::BESTSOURCE => "json",
+      ChunkMethod::BESTSOURCE => "bsindex",
       _ => return Err(anyhow!("invalid chunk method")),
     }
   )))?;
```

Uranite:
I can confirm that this PR fixes bestsource performance, my proposed changes is not necessary, but it's up to you if you want

shssoichiro:
Seems like a good idea to include anyway. Unfortunately what is blocking this PR is that something seems broken with the `select` chunk method, unrelated to this change, and I haven't been able to track down what is causing it yet.

shssoichiro:
@master-of-zen This one should be ready as well now.

PR #864, opened 2024-08-10
Fix everything breaking the CI

shssoichiro:
- Update all dependencies to their latest version
- Fix new clippy lints from the latest rustc
- Fix an issue with the `select` chunk method introduced in ffmpeg 7
  - The `SETPTS` seems unnecessary in all versions of ffmpeg, but in ffmpeg 7 it causes the filter to cut out the last several frames of the chunk

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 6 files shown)


master-of-zen (approved):
amazing) ❤️

PR #865, opened 2024-08-16
Update vapoursynth in windows-build.yml

Uranite:
I personally do encodes with .vpy script input and I haven't encountered any problem


Uranite:
Wait, turns out we can safely remove python from the github workflow file, and we can also update NASM while we're at it

PR #866, opened 2024-08-16
Update NASM and VapourSynth in windows-build.yml

Uranite:
And remove unnecessary Python setup

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -13,20 +13,14 @@ jobs:
 
       ffmpeg_ver: "7.0"
       ffmpeg_path: "C:/ffmpeg"
-      vsynth_ver: "R66"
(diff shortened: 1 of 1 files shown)


shssoichiro (approved):
lgtm, thanks!

FreezyLemon:
IIRC this was necessary because the default python version on the runner was not compatible with the chosen VS version. It's likely this changed, but note that this could break down the line when the github action runners are updated.

Uranite:
> IIRC this was necessary because the default python version on the runner was not compatible with the chosen VS version. It's likely this changed (which is why it works now), but note that this could break down the line when the github action runners are updated.

Hmm, that cant be it, since VapourSynth is portable and to compile Av1an with portable VapourSynth you dont need python for that

PR #867, opened 2024-08-16
Update compiling.md

Uranite:
Rephrase VapourSynth setup to prevent people into thinking that a certain version of VapourSynth is required
Avoid directly linking NASM for the same reason


shssoichiro (approved):
Makes sense. Thanks!

PR #873, opened 2024-08-27
Allow more than 1 worker per thread pool set by set-thread-affinity

damster101:


av1an-core/src/broker.rs
```diff
@@ -109,18 +109,10 @@ impl<'a> Broker<'a> {
 
       cfg_if! {
         if #[cfg(any(target_os = "linux", target_os = "windows"))] {
+          let available_threads = available_parallelism().expect("Unrecoverable: Failed to get thread count").get();
```
FreezyLemon, line 112: Maybe move this to the second `#[cfg]`-gated block. If it's here, this will always panic on any linux/windows platform that does not support `available_parallelism` or for a OS user without the necessary permissions on Linux.

If you move it down, it won't call `available_parallelism` unless the user sets a non-zero thread affinity.

FreezyLemon (approved):

shssoichiro (approved):
Thanks!

PR #874, opened 2024-08-27
Remove once_cell dependency

FreezyLemon:
[`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) was stabilized in 1.70 (MSRV is also 1.70).


FreezyLemon:
Oh, I missed that there's still some usages of `once_cell::sync::Lazy` in av1an-core. Those should wait for MSRV >= 1.80 where `LazyLock` is stabilized.

Issue #824, opened 2024-04-23
--vmaf does not work without target-quality


luigi311:
Looks like vmaf plot was redesigned at some point with a dependence on target-quality so its not possible to use --vmaf without --target-quality. If you add --vmaf it simply does nothing.

lyj0309:
+1

damster101:
Should be fixed now.

PR #881, opened 2024-09-25
Make --vmaf parameter work independently again

damster101:
Solves issue #824

### DIFF
--- a/av1an-core/src/context.rs
+++ b/av1an-core/src/context.rs
@@ -9,6 +9,7 @@ use std::path::{Path, PathBuf};
 use std::process::{exit, Command, Stdio};
 use std::sync::atomic::{self, AtomicBool, AtomicUsize};
 use std::sync::{mpsc, Arc};
+use std::thread::available_parallelism;
 use std::{cmp, fs, iter, thread};
 
 use ansi_term::{Color, Style};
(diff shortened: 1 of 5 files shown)


shssoichiro:
Needs a rustfmt please

damster101:
> Needs a rustfmt please

Better now?

But someone really should run rustfmt on the entire project, I had to "fix" all those indentations to match the rest of the project...

shssoichiro:
That's weird, it should already be correct for the rest of the project. The CI validates it on each MR.

damster101:
> That's weird, it should already be correct for the rest of the project. The CI validates it on each MR.

It seems rustfmt behaves VERY differently when I run it on the whole context.rs file compared to just the isolated function. I did the latter and got a ton of changes everywhere, especially lots of two space indentations were increased in size to four spaces.
But, if I run rustfmt on the whole context.rs file, it seems to recognize that a different style is used and doesn't perform any of these changes.

Btw, I would have expected that this behavior could be influenced by the user for whole project files, but there doesn't seem to be an option for it...

damster101:
So, it looks like the way to get modern Rust formatting for a large project is to omit all the lines starting with `use` at the beginning of the Rust project files when formatting them with rustfmt.

damster101:
Once some part of the file is formatted in the modern default style, running rustfmt on the whole file will change everything to it.

shssoichiro (approved):

PR #886, opened 2024-10-30
Add Richly-Typed loadscript VapourSynth Script to Source Control

BoatsMcGee:
The current way Av1an writes vpy files is prone to errors from the developer side. Currently, the Rust source code itself contains a separate string for each "import" method which is tailor-made for that method. Most of the strings are identical between methods and are needlessly repetitive. More crucially, these strings are not recognized as Python by the IDE because understandably the IDE isn't expecting to encounter a different language in a random string. Due to this limitation, developers won't appreciate rich typings and intellisense the IDE would offer should that Python code be written where it is expected: in a .py file. This leaves contributors prone to making mistakes they might otherwise avoid should they have the guidance modern IDEs provide. This is easily solved by designing and writing the python file directly into the project files structure and having that source-controlled.

## Moving Python Into a Dedicated File

This is a proposal to use the [`include_str!`](https://doc.rust-lang.org/std/macro.include_str.html) macro which can include an entire text file, or in this case a Python script, directly into the built executable. Essentially, we can move Python code out of .rs files and into their respective .py or .vpy files and use them however we want in the Rust code. In this PR, I unified the scripts for each import method into 1 .vpy file and modified the string itself for each method to set variables for the script to access during runtime. We can also avoid modifying the text altogether by instead providing the values through environment variables when the script is invoked. This simplifies the embedded Python code and the Rust code all while making it more palatable for potential contributors.

## Enhancements

This change also allows us to leverage the Python runtime as seen in the `bestsource` case when different versions produce different outputs. In short, the recent release of [R8](https://github.com/vapoursynth/bestsource/releases/tag/R8) allows setting the cache file path to an absolute path where previous versions did not allow this whatsoever. For a fuller picture, see the comments in the .vpy script.

This suggestion falls outside the scope of this PR but the .vpy script can be further modified to support a potential feature where users can specify a preferred order of chunking methods to attempt before the script fails. For instance, users may prefer DgDecNV, BestSource, and LSmashWorks in that order and the script can fallback to less preferred methods should previous ones fail or not be installed.

## Considerations

I have very limited experience with Rust currently and the `create_vs_file` function could be further improved from my introduced changes so please feel free to make or suggest necessary tweaks.

Thank you,
\- Boats


shssoichiro:
This looks interesting and I like the idea. I'll take a look through the implementation most likely later today.

shssoichiro (approved):
I think this is good.

shssoichiro:
I spoke too soon, looks like it needs a `cargo fmt` real quick.

BoatsMcGee:
Sorry about that, I assume it's just a style issue and not anything related to the logic, right? Having run `cargo fmt` locally the only related change I see is line 254 in `vapoursynth.rs` was collapsed into a single line which is totally fine by me. However, I also see that `util.rs` and `scenes.rs` also changed in the imports. Everything seems harmless so I'll go ahead and commit that.

BoatsMcGee:
Is there anything else that needs to be done? I synced the fork for the latest changes and made sure that there's no other formatting issues.

Thanks,
\- Boats

shssoichiro:
It seems like there's some issue going on where the tests are failing. see https://github.com/master-of-zen/Av1an/actions/runs/11674361550/job/32550276127?pr=886

I may have time to help look into this later today, been quite busy lately but I'll see what I can figure out.

BoatsMcGee:
This last commit expands on this PR to resolve Issue #931 using my preferred solution presented in my comment there. Briefly, this solution adds a code path specifically for Scene Detection to the generated `loadscript.vpy` and is only enabled by passing an environment variable to the process. To support downscaling on user-provided custom VapourSynth scripts, I implemented the copy and modify/inject method to replace the `VIDEO_NODE.set_output(` with a default Bicubic scaling function before outputting.

This is sufficient for most users but it won't work for scripts that use third party libraries that abstract the `.set_output()` function. A better solution would be to use the vapoursynth-rs crate to programmatically access the first output video node, downscale, and then output to stdout for the y4mDecoder or perhaps directly with VapourSynthDecoder. However, I'm not familiar enough with that library to implement that solution quickly.

Thanks,
\- Boats

BoatsMcGee:
I fixed the error related to my changes but I don't understand why a completely different part of the code is suddenly causing an error in all-tests. I also don't see these errors when building/checking locally (with `--release`) so I may need some help here.

shssoichiro:
We've been seeing the same error across all open PRs. I was able to fix one test failure, and then another one appeared right behind it. Unfortunately I have been busy with other tasks so I haven't been able to focus on fixing it, but I really need to as it's blocking several open PRs right now...

BoatsMcGee:
I'll try to hunt some errors later to get the needle moving along. Looks like the test uses `cargo clippy --profile ci -- -D warnings` so I think I can avoid spamming GitHub Actions with commits when troubleshooting that. Docs indicate that command will fail when there is any warning whatsoever so that explains why the check/build commands don't indicate anything.

BoatsMcGee:
I think this last commit should get us all those green checkmarks.

I fixed the issues with this code calling `to_absolute_path(...)` on paths that may not exist yet which was the root cause of the "invalid file or directory" errors. PR #924 should be expanded to the relevant `format!()` calls in `create_vs_file(...)` so that path formatting is more robust.

I reverted the clippy suggestion commmits once I noticed PR #935 and merged that instead. I did notice that one of the tests fails for Windows environments since the test assumes Linux paths.

```log
test chunk::tests::test_chunk_output ... FAILED // This fails because the test is written for Linux/UNIX paths, not Windows
test target_quality::tests::test_bisect ... FAILED

---- chunk::tests::test_chunk_output stdout ----

thread 'chunk::tests::test_chunk_output' panicked at av1an-core\src\chunk.rs:164:5:
assertion `left == right` failed
  left: "d/encode/00001.ivf"
 right: "d\\encode\\00001.ivf"
```

While I would very much like to be done with this PR and move on to a different issue, I suggest that we (eventually) completely obsolete the loadscript files and entirely use the vapoursynth-rs crate instead. I can make a new PR for that discussion if you or anyone else finds that worthwhile.

---

Below is my poor attempt at debugging the issue assuming `include_str!()` was the cause of the issue. I've kept it because there's some details and alternatives that could be useful in the future in case `include_str!()` is unreliable in any way. *No need to read it.*

There are other issues. The `include_str!()` macro fails exclusively in the container and also the CI workflow that uses the container. I don't fully understand it but it doesn't fail during compile time when it "expands" the macro to "store" the file that the string (relative path) points to. It fails during runtime and leads to the `Error: No such file or directory (os error 2)` when running the tests - hence the tests failing despite the above fix.

At first, I thought that the path was incorrect because the environment is different but `include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/src/loadscript.vpy"))` and `include_str!(env!("CARGO_MANIFEST_PATH"))` also yield the same error. "CARGO_MANIFEST_PATH" at the very least should exist during compile time since Cargo sets it so I don't understand why the executable built in the container or in CI fails to load any file.

After throwing a lot of pasta at the wall I've come up with a hacky solution that seems to work on my machine, docker, and in CI:

1. Create `av1an-core/build.rs`
2. Retrieve text from ~~`include_str!("src/loadscript.vpy")`~~ `std::fs::read_to_string("src/loadscript.vpy")`
3. Encode text in Base64
4. Store Base64 string in compile-time environment variable with [`println!("cargo:rustc-env=LOADSCRIPT={}", encoded_str);`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env)
```rs
use base64::Engine as _;

fn main() {
  use std::path::Path;

  if !Path::new("src/loadscript.vpy").exists() {
    panic!("File src/loadscript.vpy does not exist");
  }

  let loadscript_template =
    std::fs::read_to_string("src/loadscript.vpy").expect("Failed to read src/loadscript.vpy");

  // Encode the multiline string into a single line
  let encoded_str = base64::engine::general_purpose::STANDARD.encode(&loadscript_template);

  // Store the encoded string in an environment variable
  println!("cargo:rustc-env=LOADSCRIPT={}", encoded_str);

  // Retrieve the string from the environment variable (runtime only)
  // let retrieved_str = std::env::var("LOADSCRIPT").unwrap();
  let retrieved_str = encoded_str;

  // Decode the string back to the original multiline string
  let deserialized_loadscript_template = base64::engine::general_purpose::STANDARD
    .decode(retrieved_str)
    .unwrap()
    .iter()
    .map(|&c| c as char)
    .collect::<String>();

  // Assert that the retrieved string matches the original
  assert_eq!(deserialized_loadscript_template, loadscript_template);
}
```
5. Retrieve Base64 string during runtime with `env!("LOADSCRIPT")` macro
6. Decode Base64 string
```rs
let loadscript_template = base64::engine::general_purpose::STANDARD
    .decode(lsc)
    .unwrap()
    .iter()
    .map(|&c| c as char)
    .collect::<String>();
```

shssoichiro:
Awesome. Okay, I'm going to go through the smaller MRs first, rebase and merge them, then rebase and review this one.

av1an-core/src/scene_detect.rs
```diff
@@ -242,26 +242,16 @@ fn build_decoder(
...
-            .args(filters)
-            .arg("-")
-            .stdout(Stdio::piped())
-            .stderr(Stdio::null())
-            .spawn()?
-            .stdout
-            .unwrap(),
-        )?)
+
+        Decoder::Y4m(y4m::Decoder::new(command.spawn()?.stdout.unwrap())?)
```
shssoichiro, line 254: Oh. Yeah, we don't need ffmpeg at all here because we can do the resizing inside of Vapoursynth now. Nice.

av1an-core/src/vapoursynth.rs
```diff
@@ -210,59 +213,105 @@ pub fn create_vs_file(
...
-        cache_file
-      )
-      .as_bytes(),
-    )?;
+          ChunkMethod::DGDECNV => dgindex_path.display(),
+          _ => source.display(),
+        }
+      ),
+    )
+    .replace(
```
shssoichiro, line 257: I'm curious, can we do this by setting the respective environment variables instead of by replacing each of these in the text?

av1an-core/src/vapoursynth.rs
```diff
@@ -210,59 +213,105 @@ pub fn create_vs_file(
...
+  source: &Path,
+  downscale_height: Option<usize>,
+) -> anyhow::Result<PathBuf> {
+  let temp: &Path = temp.as_ref();
+  let scd_script_path = temp.join("split").join("scene_detection.vpy");
+  let mut scd_script = File::create(&scd_script_path)?;
+
+  let source_script = std::fs::read_to_string(source)?;
+  if let Some(downscale_height) = downscale_height {
+    let regex = Regex::new(r"(\w+).set_output\(").unwrap();
```
shssoichiro, line 299: Likewise, this regex and replacement may not be necessary at all if we can do it using the loadscript + env vars.

shssoichiro: Never mind, I see now why we're doing this here.

PR #893, opened 2024-11-23
Fix compatibility with x265 4.0

shssoichiro:
x265 version 4.0 introduced an undocumented breaking change to the CLI where passing the input without the `--input` flag is no longer supported. Therefore, we need to update av1an accordingly. The `--input` syntax is also supported by x265 3.x, so this change is backwards compatible.


master-of-zen:
Looks good to me, but target quality needs fixing.
I will merge for now so we can move on

shssoichiro:
Yeah not sure what's up with target quality there, I am not sure why that would need a different fix than the regular encode CLI

Issue #907, opened 2025-01-14
x265 issue, problem with muxing


raysv:
Hi, 
Trying to do some 265 encodes with 0.4.4.

However running into this problem:

thread '<unnamed>' panicked at /home/raysv/.cargo/registry/src/index.crates.io-6f17d22bba15001f/av1an-core-0.4.4/src/broker.rs:172:49:
called `Result::unwrap()` on an `Err` value: EncoderCrash { exit_status: ExitStatus(unix_wait_status(256)), stdout: , stderr: , source_pipe_stderr: [vost#0:0/wrapped_avframe @ 0x5566be4c0a40] Error submitting a packet to the muxer: Broken pipe
    Last message repeated 1 times
[out#0/yuv4mpegpipe @ 0x5566be4c02c0] Error muxing a packet
[out#0/yuv4mpegpipe @ 0x5566be4c02c0] Task finished with error code: -32 (Broken pipe)
[out#0/yuv4mpegpipe @ 0x5566be4c02c0] Terminating thread with return code -32 (Broken pipe)
[out#0/yuv4mpegpipe @ 0x5566be4c02c0] Error writing trailer: Broken pipe
[out#0/yuv4mpegpipe @ 0x5566be4c02c0] Error closing file: Broken pipe
, ffmpeg_pipe_stderr: None }

Anyone got any suggestions on what might be wrong here? svt-av1 works just fine.

SwareJonge:
Looks like no inputs are mapped for x265 in target quality mode, haven't tested if adding it fixes it.
See encoder.rs at line [753](https://github.com/master-of-zen/Av1an/blob/master/av1an-core/src/encoder.rs#L753-L764) and [794](https://github.com/master-of-zen/Av1an/blob/master/av1an-core/src/encoder.rs#L794-L802)

PR #910, opened 2025-01-26
Fix encoding with multiple encoders

SwareJonge:
Currently av1an can't use multiple encoders at the same time when specified in the zone overrides as it always passes the args from the command line when creating `chunks.json`. 
With this fix it results in the expected behaviour, since according to the example in ``av1an --help`` it should be possible to do this already.
I'm not that familiar with Rust so if there are things i could've done better here, please point it out.


SwareJonge:
Well, that should be it. Not sure what causes the current error but that's out of my control, looks like it's the same error as issue #907

shssoichiro:
The failing test seems to have been broken in master for a while, unfortunately I haven't been able to track down what the issue is. I should probably skip those tests until we find a fix (not ideal, but it is what it is).

shssoichiro (approved):

shssoichiro:
@master-of-zen Would you mind merging this? I believe you're the only one with the ability to override the CI. The failures are not related to this MR, it's the same failure from master that we haven't been able to figure out the cause of.

PR #912, opened 2025-01-28
Update dependencies and fix lints

### LINKED ISSUE #911: FFmpeg 7.1 support
`ffmpeg-the-third` should be updated to `v3.0.0+ffmpeg-7.1` to ensure proper support for FFmpeg 7.1, even though it currently works fine.

The real reason I created this GitHub issue is that I'm trying to compile Av1an on Termux, and I believe updating ffmpeg-the-third to `v3.0.0+ffmpeg-7.1` might fix the issue. It would be awesome to have Av1an running on Termux.

This will require additional changes, as I've tried just updating the version of `ffmpeg-the-third`, but it fails to compile. It will also require updating `av-scenechange`'s `ffmpeg-the-third` to `v3.0.0+ffmpeg-7.1`.

shssoichiro:
Updates all dependencies to the latest versions,
except for nom, which requires some more code changes.

Closes #911

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


shssoichiro:
The test failure seems like an outstanding issue with the target-quality tests from `master`.

Issue #921, opened 2025-02-09
Fix the target quality tests


shssoichiro:
The target quality tests seem to be failing with an encoder crash, even though there are no reports of this happening to real users. We should fix these tests.

I will temporarily disable them so that they no longer block every MR coming in, but once they are fixed we should re-enable them.

PR #922, opened 2025-02-09
Temporarily disable target quality tests

shssoichiro:
Related to #921


shssoichiro:
I tried to get a proper fix in https://github.com/master-of-zen/Av1an/pull/923 but it was not as simple as I hoped. So I'm going to open this up for now until we can fix the issue, since the tests have been broken in `master` for a while.

shssoichiro:
Nah, I'm going to go ahead and try to find a proper fix for this. We are seeing more users reporting the issue, so I think just disabling the tests is irresponsible.

PR #923, opened 2025-02-09
Fix x265 target quality

shssoichiro:
Closes #907

PR #927, opened 2025-02-24
Fix Scene Detection frames mismatch causes invalid frame access crashes

BoatsMcGee:
Scene Detection currently uses FFmpeg, even when the chunking method used later encodes the video with a VapourSynth script. This leads to a mismatch in the frame count and crashes on the offending chunk. Scene Detection should instead use the same decoding method as Av1an does when encoding chunks in order to avoid this mismatch.

Additionally, when testing this fix I discovered a race condition that fails scene detection when retrieving the frame count in a thread. I don't fully understand why this occurs but moving the resolution of that thread *before* scene detection starts resolves that race condition.

This issue can be replicated by improperly cutting a lossless clip and using that as the input while specifying a chunking method that generates/uses a VapourSynth script.

As a workaround, users can instead input a VapourSynth script instead of the clip itself and the Scene Detection should perform the same function as this PR.

Thank you,
\- Boats M.

### DIFF
--- a/av1an-core/src/context.rs
+++ b/av1an-core/src/context.rs
@@ -120,7 +120,7 @@ impl Av1anContext {
 
       // frames need to be recalculated in this case
       if self.frames == 0 {
-        self.frames = self.args.input.frames()?;
+        self.frames = self.args.input.frames(self.vs_script.clone())?;
         done.frames.store(self.frames, atomic::Ordering::Relaxed);
       }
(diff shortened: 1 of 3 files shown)


shssoichiro (approved):
Thank you, this seems like it makes sense. This will need a rebase after #923 is merged to fix the CI, but other than that it looks good.

PR #935, opened 2025-04-17
Fix the CI

shssoichiro:


shssoichiro:
@master-of-zen Would you mind approving? I ran the test suite on this MR twice just to make sure it was not a fluke, and it passed both times, as strange as that is.

master-of-zen (approved):

PR #960, opened 2025-04-23
See if lcov gives more intuitive results

shssoichiro:


shssoichiro:
Yeah, this is better. There was some weirdness with the branch-based coverage that showed partial coverage on lines where that didn't make sense.

PR #962, opened 2025-04-24
Improve CLI documentation and formatting

BoatsMcGee:
Rather than just pasting the CLI help output, I formatted the general documentation to be a bit more like a browser page that is uniquely useful from the CLI output. At the top is a table of the arguments with each name linking to its respective section below with more details and examples as necessary. There's a couple of additional details added mostly for document consistency so it may need some adjustments.

I have only done the general page just as a starter to get some improvements in the Book so I'm very much open to suggestions. Once we agree on a solid preference for the format I can do the rest.

Thank you,
\- Boats M.


shssoichiro (approved):
Thanks, this looks nice

PR #964, opened 2025-04-25
Improve CLI Documentation formatting

BoatsMcGee:
This is a continuation of #962 for the remaining CLI documents. There are also fixes to the main readme.md - mainly updates to links and some grammar.

Next steps:

* Update inaccurate information on some of the chunking methods - docs suggest VS methods don't generate temporary files but they do (mainly indexing files)
* Update `av1an/src/main.rs` parameters to match the new documentation
  * Reorder for consistency
  * Update grammar (ex. "ffmpeg" must be "FFmpeg" to comply with license [checklist item 15](https://www.ffmpeg.org/legal.html#:~:text=Do%20not%20misspell%20FFmpeg%20(two%20capitals%20F%20and%20lowercase%20%22mpeg%22).))
* Make new pages to clean up main readme.md:
  * Installation for more details on installing - doing this in readme would be too verbose
  * Usage + examples - would be nice to have animated webp showcase of the TUI (both on readme and docs)
* Developer documentation - Each file/configuration generated or used by Av1an should be documented: Why do we need this, what does it do, etc. and most importantly *how should it be implemented*
  * Loadscript.vpy
  * Scenes.json
  * Chunks.json
  * Done.json

I would also like to further improve information on VMAF and Target Quality. I invite anyone more familiar than I am to fill in missing info. I don't use these features so I was a bit lazy on some of the sections like `Target Quality -> Minimum Quantizer`. I would have to further dig into the code to get exact values but I'm also more interested in reimplementing it differently with newer features and procedures.

Thanks,
\- Boats M.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,14 @@
 
 <a href="https://www.buymeacoffee.com/master_of_zen" target="_blank"><img src="https://cdn.buymeacoffee.com/buttons/v2/default-yellow.png" alt="Buy Me A Coffee" style="height: 60px !important;width: 217px !important;" ></a>
 
-Av1an is a video encoding framework. It can increase your encoding speed and improve cpu utilization by running multiple encoder processes in parallel. Target quality, VMAF plotting, and more, available to take advantage for video encoding.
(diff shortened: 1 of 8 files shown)


shssoichiro:
Would you mind running `cargo fmt` on the branch, please? Looks like that is the cause of the CI failure. Thanks.

shssoichiro (approved):

Issue #936, opened 2025-04-17
Suggestion: Allow encoder to finish chunk when cancelling


plshelpidkwhatimdoing:
I don't know if it's possible, but could you make it so when `ctrl+c` is pressed one time, the encoder is allowed to finish the chunk it's doing but not start another one. Pressing `ctrl+c` 2 - 3 times in quick succession can force cancel (like how cancelling is done currently). That way stopping can be done with minimal loss if you're aren't in a hurry to end it

Issue #597, opened 2022-03-19
Feature: graceful/soft close option


iamtherobin:
Would it be possible to implement a soft close command to av1an which allows the current chunks to finish and mark them done before closing? Would be useful when I don't immediately need the resources back and also don't want the time spent on the current chunks wasted.

dexeonify:
I wonder what use case is this useful for. If I notice a problem with the encoded video or I need the system resources, I would terminate av1an and all of its processes immediately.

In the meantime, you can use the <kbd>Pause/Break</kbd> key (on Windows, <kbd>Ctrl</kbd>+<kbd>Z</kbd> on Linux) to temporarily pause av1an from piping new frames. This will let the encoders finish whatever frames its currently encoding, then exit gracefully. It will only free up your CPU though, not RAM.

n00mkrad:
+1 for OP's request, if I want to pause an encode (and free up the RAM it takes) it kinda sucks that you lose the current chunk progress. Usually I'm not in a hurry so letting the current workers/chunks finish would be a nice thing.

PR #966, opened 2025-04-26
Add CTRL+C SIGINT handling

BoatsMcGee:
Sending `SIGINT` (pressing CTRL+C) while encoding currently terminates the chunk encoder, causing Av1an to retry the chunk and requiring the user to send multiple signals to terminate. This PR aims to bring that down to 1 command. There are also 2 open issues requesting that sending `SIGINT` should let the chunk(s) finish but not start any more until all are completed and Av1an exits early with no progress lost. That is the secondary goal for this PR but I'm not sure how to implement that yet.

This is a work in progress and there needs to be better logging for these new scenarios. It also needs to be applied in more parts of the code, such as in the Target Quality search before chunk encoding starts. I also need to test several different configurations like multiple input files, multiple workers, Linux, etc.

Thanks,
\- Boats M.

### DIFF
--- a/av1an-core/src/broker.rs
+++ b/av1an-core/src/broker.rs
@@ -4,7 +4,7 @@ use std::io::Write;
 use std::path::Path;
 use std::process::ExitStatus;
 use std::sync::{
-  atomic::{AtomicBool, Ordering},
+  atomic::{AtomicU8, Ordering},
   mpsc::Sender,
   Arc,
 };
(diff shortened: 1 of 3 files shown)


BoatsMcGee:
This last commit is much more refined. Each CTRL+C (`SIGINT`) signal increments an atomic counter. The first signal allows for the current set of workers to finish their task and update the `done.json` appropriately. It also logs "Shutting down. Waiting for current workers to finish...". Sending another `SIGINT` signal will log "Shutting down all workers..." and no workers will be permitted to finish encoding.

However, there's a big caveat to this and it affects how the immediate shutdown works; CTRL+C also terminates the `tokio::Command` spawns so all workers will restart after the first `SIGINT` signal. But the 2nd signal takes advantage of that bug to assess whether or not the worker should restart if CTRL+C was sent twice.

So this commit addresses Issues #936 and #597 partially - it's not as graceful because there's no intercept for signals to the `tokio::Command` so all workers get terminated every time CTRL+C is pressed. If someone could figure out how to prevent that then that would be great.

Still need to test in my UNIX environment, with multiple inputs files, and during Target Quality searching. For now I'll remove the draft as this PR is now stable and achieves its primary objective - preventing zombie workers from lingering and repeated CTRL+C signals leading to orphaned encoder processes.

av1an-core/src/broker.rs
```diff
@@ -109,9 +113,22 @@ impl Broker<'_> {
       drop(sender);
 
       crossbeam_utils::thread::scope(|s| {
+        let terminations_requested = Arc::new(AtomicU8::new(0));
+        let terminations_requested_clone = terminations_requested.clone();
+        // let active_workers = Arc::new(Mutex::new(Vec::new()));
```
shssoichiro, line 118: Would prefer not to leave commented code in, please.

BoatsMcGee: Ah, sorry about that - must've accidentally staged that while separating out test code for the next phase: target quality. Once I figure that out that next commit will already have that removed.

shssoichiro:
Thanks, I think this is good for a starting point. 🎉

BoatsMcGee:
I couldn't figure out a good solution for passing `terminations_requested` to the `per_shot_target_quality_routine(..)` without having to refactor a bunch of other functions so I don't have a clean solution for that.

I'll try to make another PR that attempts to print out the Target Quality info in the mini progress bar and I'll make sure that has the stray comment removed. That PR should address at least one of the open issues and should provide some reassuring feedback to the user.

PR #977, opened 2025-04-28
Move tracing instrumentation to debug

shssoichiro:
I found that the big blobs of debug info that were being printed on certain log lines, e.g. the resume logs, were a result of `tracing::instrument`. This change moves those all to debug level, so they are not printed to the console by default. This also cleans up a few other minor items, which are not necessary, but I already cleaned them up while debugging so I figured they are worth keeping.


master-of-zen (approved):
LGTM

PR #979, opened 2025-04-29
Set the output FPS when concatenating

### LINKED ISSUE #475: FPS of the encoded video is wrong
Hi! I encode a bunch of videos which video FPS is 24, but the encoded ones have FPS 24.122. The only way to change it back is by using MKVToolNix GUI, which is very inconvenient.

The command I am using is following:

`docker run --privileged -v /REDACTED:/videos --user 1000:1000 -it --rm masterofzen/av1an:latest -i a.mkv -s mats -l logfile -o a30.mkv -e aom -c mkvmerge -a ' -an ' -v ' --cpu-used=4 --end-usage=q --cq-level=30  --threads=4 --arnr-strength=4 --bit-depth=10 --enable-fwd-kf=1 --lag-in-frames=35 --max-partition-size=64 --enable-qm=1 --enable-chroma-deltaq=1 --quant-b-adapt=1 --tile-columns=1 --tile-rows=0 --enable-dnl-denoising=0 --denoise-noise-level=8 ' --vmaf --resume`

shssoichiro:
This ensures that certain encoders such as vpx which do not properly set
    the frame rate on their outputs do not cause av1an to output videos with
    the wrong frame rate. This only works with mkvmerge concat method.
    As a result of this plus the existing issue with ffmpeg not
    concatenating x265 outputs correctly, this also changes the default
    concat method to mkvmerge.

Fixes #475


shssoichiro:
@master-of-zen What are your thoughts about the change on the default concat method? My biggest concern is that mkvmerge only supports mkv (and maybe webm? but definitely not mp4). But ffmpeg has too many annoyances that make it not optimal for concatenating and muxing--it wasn't even possible to set a framerate on a video when concatenating without reencoding.

shssoichiro:
I am going to go ahead and pull the trigger on this. It's been a repeated issue for VP9 users in particular, and I think the breaking change of the default option changing is for the best.

Users who want the old behavior can use `--concat ffmpeg`.

PR #982, opened 2025-05-03
Only apply specific default params if the user specifies encoder params

### LINKED ISSUE #981: Default encoding parameters `--preset 4 --keyint 0 --scd 0 --rc 0 --crf 25` are forcefully added and cannot be removed.
In the latest git version of av1an, when `--encoder svt-av1` is specifed, encoding parameters `--preset 4 --keyint 0 --scd 0 --rc 0 --crf 25` are always added even if the user manually specify their own `--video-params` and cannot be removed in any manner.  
For example, this makes it completely impossible to do VBR or CBR encodings. Even if you specify `--rc 1` in the `--video-params`, the `--crf` parameter still set it to a CRF encoding.  

It's good to have a default parameters when the user doesn't specify `--video-params`, but please hand the control completely to the user when the user provides their own`--video-params`.

shssoichiro:
Closes #981

### DIFF
--- a/av1an-core/src/encoder.rs
+++ b/av1an-core/src/encoder.rs
@@ -308,7 +308,8 @@ impl Encoder {
     }
   }
 
-  /// Returns default settings for the encoder
+  /// Returns default settings for the encoder.
+  /// These will be excluded if the user specifies any custom params.
   pub fn get_default_arguments(self, (cols, rows): (u32, u32)) -> Vec<String> {
     /// Integer log base 2
     pub const fn ilog2(x: u32) -> u32 {
@@ -332,8 +333,6 @@ impl Encoder {
           "--cpu-used=6",
           "--end-usage=q",
           "--cq-level=30",
-          "--disable-kf",
-          "--kf-max-dist=9999"
         ];
 
         if cols > 1 || rows > 1 {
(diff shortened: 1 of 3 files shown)


Akatmks:
Thank you! I saw `args.no_defaults` in the diff, and I think that's exactly what I'd like.  
As long as `--no-default` option is presented, I'm not against av1an inserting its own parameters at all. I think it is great that I don't have to write `--scd 0` and `--keyint -1` manually and av1an will take care of it.

Thanks!

shssoichiro:
Thanks for the feedback. I will change it to keep the `--no-defaults` option then, and have that remove everything including the additional keyframe args, although disabling those will not be recommended.

Akatmks:
Sorry! I wasn't paying attention. Av1an actually already has `--no-defaults` (and av1an documentation has also started working again since last time I checked)!
Had I noticed the `--no-defaults` option I would not create the Issue. Sorry for that!

---

Btw I want to ask another unrelated question. Sorry for it. Is av1an's scenes file considered an internal file? In the sense that the format for zones file are clearly defined and would not change unless it has to. Is it the same for scenes file as well, or is it considered an interal format that can be changed frequently to meet any new features in new versions?

shssoichiro:
The scenes.json file is not likely to change, although it is not officially defined as a stable spec, so it _may_ change at any point. i.e. yes, it is an internal file at the moment. Though if there is value in it, we could consider defining a stable spec for it.

Akatmks:
Thanks! I understand that you are not paid developing av1an and I'm not urging you of anything if I ever sound like that. The only reason I'm [using](https://github.com/Akatmks/Akatsumekusa-Encoding-Scripts/blob/e53b44dba03a89f9dd6ac6ee8c95090c62d5bf52/Progression-Boost/Progression-Boost.py#L33) scenes file is because zones file is almost completly unusable at the moment <sup><sub>(there are issues with `--scene-detection av-scenechange` using zones file as well; just I haven't found out in which exact situation it happens so I haven't reported them)</sub></sup>. If it is a semi-stable format I think I can deal with it.  
If I have time maybe I will one day trying to help fix some issues as well.  
Thank you so much for maintaining av1an.

shssoichiro:
Closing as the `--no-defaults` flag is found to be sufficient.

Issue #1009, opened 2025-06-04
[Feature Request] Make Temporal Weighting in VMAF Togglable


GreatValueCreamSoda:
A common issue with VMAF is during scenes of high motion, that are usually lower scores, VMAF will artificially boost scores to 100. This is due to VMAFs primitive motion compensation which is weighted on a SAD difference of the frames blurred with a Gaussian. 

Due to this certain content like for example anime can be unfairly weighted as looking better than it actually is within frames of motion. 

Making this a togglable option would help significantly with specific types of content.

PR #1012, opened 2025-06-09
Add --percentile, --alt-vmaf & Better Search

### LINKED ISSUE #1008: [feature Request] Add multiple averaging methods for target quality encoding
A long standing issue with Av1an has been the inability to control the averaging methods used for per frame statistics from VMAF and now SSIMULACRA2. 

As it stands right now, VMAF is hard coded to 1% lows, while the PR trying to add support for SSIMULACRA2 is hard coded to use arithmetic mean. 

For particular content like slice of life anime arithmetic mean can be inaccurate to the final encodes real quality due to encoders prioritizing static content within frames and heavily penalizing movements at higher CRF levels. These dips in quality not being accurately reflected in the mean of all frames. 

With the advent of new metrics being added to Av1an, and assuming VMAF is fixed, users will now be able to target high quality or “transparent” encodes using more psyco-visually aware metrics. Eliminating the requirement of using extremely low CRFs to achieve transparency throughout the video, and instead only using it where needed. For these transparent encodes, usually looking at a 5/10% lows or absolute minimum frame score is preferable. Not only for the aforementioned issues, but to guarantee a set "lowest" quality. 

I have personally seen 5/10% lowest frames be recommended by (and I personally endorse) by ex members of the svt-av1-psy team. It is common for users in their own scripts to use geometric or harmonic mean. Min as mentioned is desirable by some users. 

Some metrics have inverse scaling where instead of higher is better, lower is better. Said cases do need to be handled separately.

emrakyz:
Fixes #1008, #1009

This PR attempts to add **2 new CLI flags** and improve/change the CRF search algorithm by implementing Catmull-Rom Spline Interpolation, effectively making CRF search more accurate and faster and preventing its unwanted early-exit behavior.

```
--alt-vmaf --percentile 0.1 --target-quality 90 --probes 99 --probing-rate 1 --probe-slow --min-q 1 --max-q 70
```

Some parts might need to change or be refactored after #988 gets implemented.

It also adds an alternative VMAF calculation which disables the motion compensation that artifically boosts scores. This method also relies on perceptual YUV weighthing (`4 + 1 + 1 / 6`). 

- Start the search loop

# Phase 1 (No History)

# Phase 2 (two data points)

# Phase 3: Three+ Data

# Conditions

# Termination

# Selection

# Extra Probing Features

# Added CLI Flags
- `--alt-vmaf` : 
Disable motion --> Perceptually weight the scores --> Sort them based on given percentile --> Use the percentile result as the value
- `--percentile`: Use any given percentile as an averaging method.

image

### DIFF
--- a/av1an-core/src/concat.rs
+++ b/av1an-core/src/concat.rs
@@ -172,7 +172,7 @@ pub fn mkvmerge(
         let p = p.as_ref().display().to_string();
         if let Some(path) = p.strip_prefix(UNC_PREFIX) {
             if let Some(p2) = path.strip_prefix("UNC") {
-                format!("\\{}", p2)
+                format!("\\{p2}")
             } else {
                 path.to_string()
             }
--- a/av1an-core/src/progress_bar.rs
+++ b/av1an-core/src/progress_bar.rs
@@ -42,7 +42,11 @@ fn pretty_progress_style(resume_frames: u64) -> ProgressStyle {
         .template(INDICATIF_PROGRESS_TEMPLATE)
         .unwrap()
         .with_key("fps", move |state: &ProgressState, w: &mut dyn Write| {
-            let resume_pos = state.pos() - resume_frames;
+            let resume_pos = if state.pos() < resume_frames {
+                resume_frames
+            } else {
+                state.pos() - resume_frames
+            };
             if resume_pos == 0 || state.elapsed().as_secs_f32() < f32::EPSILON {
                 write!(w, "0 fps").unwrap();
             } else {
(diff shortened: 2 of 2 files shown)


BlueSwordM:
@shssoichiro Everything LGTM. What about your side? I've already tried it on my slow laptop, and it makes quite the nice difference; a bit faster actually in some limited scenarios.

shssoichiro:
I accidentally broke it. Bad things happen when you name your remote branch "master" and I have to try to push a commit to it :cry:

emrakyz:
@shssoichiro 

Can re-open it

shssoichiro:
I can't reopen it because your master is now identical to av1an's master. And I don't have permission to push to your branch anymore since the PR isn't open. (Great logic, Github.) So I created it from a different branch since it's also not possible to change the source branch for a PR (again, thanks Github.) https://github.com/rust-av/Av1an/pull/1013

PR #1013, opened 2025-06-09
Add --percentile, --alt-vmaf & Better Search

### LINKED ISSUE #1008: [feature Request] Add multiple averaging methods for target quality encoding

shssoichiro:
Fixes #1008, #1009

This PR attempts to add **2 new CLI flags** and improve/change the CRF search algorithm by implementing Catmull-Rom Spline Interpolation, effectively making CRF search more accurate and faster and preventing its unwanted early-exit behavior.

```
--alt-vmaf --percentile 0.1 --target-quality 90 --probes 99 --probing-rate 1 --probe-slow --min-q 1 --max-q 70
```

Some parts might need to change or be refactored after #988 gets implemented.

It also adds an alternative VMAF calculation which disables the motion compensation that artifically boosts scores. This method also relies on perceptual YUV weighthing (`4 + 1 + 1 / 6`). 

- Start the search loop

# Phase 1 (No History)

# Phase 2 (two data points)

# Phase 3: Three+ Data

# Conditions

# Termination

# Selection

# Extra Probing Features

# Added CLI Flags
- `--alt-vmaf` : 
Disable motion --> Perceptually weight the scores --> Sort them based on given percentile --> Use the percentile result as the value
- `--percentile`: Use any given percentile as an averaging method.

image


shssoichiro:
Sorry @emrakyz I had to reopen this as a new one because I broke the old one. Whoops.

shssoichiro:
Okay, looks good to me. I fixed the formatting, a compiler error in the tests, and one of the many Copilot suggestions (the rest I don't really care about). Once CI passes I can merge.

shssoichiro:
It appears there is an issue with the encode tests: `target/ci/av1an -i tt_sif.y4m -e x265 --pix-format yuv420p -c mkvmerge --sc-method fast -y -o "tt_sif.mkv" -v " --preset ultrafast" --target-quality 95 --probe-slow` is resulting in an error:

```
ERROR encode_file: mkvmerge concatenation failed with output: Output {
    status: ExitStatus(
        unix_wait_status(
            512,
        ),
    ),
    stdout: "mkvmerge v91.0 ('Signs') 64-bit\nError: The file '00000.mkv' could not be opened for reading: open file error.\n",
    stderr: "",
}
command: cd ".11a56f9/encode" && "mkvmerge" "@../options.json"
Error: mkvmerge concatenation failed
```

emrakyz:
@shssoichiro 

Hmm, I haven't tested with other encoders.

Can it be related to how I want to copy the last `ivf` that is tested (eliminating extra encoding). Check this part:

```Rust
            if tq.probe_slow && chunk.tq_cq.is_some() {
                let optimal_q = chunk.tq_cq.unwrap();
                let probe_file = std::path::Path::new(&self.project.args.temp)
                    .join("split")
                    .join(format!("v_{}_{}.ivf", optimal_q, chunk.index));
                    
                if probe_file.exists() {
                    let encode_dir = std::path::Path::new(&self.project.args.temp).join("encode");
                    std::fs::create_dir_all(&encode_dir).unwrap();
                    let output_file = encode_dir.join(format!("{:05}.ivf", chunk.index));
                    std::fs::copy(&probe_file, &output_file).unwrap();
                    
                    inc_mp_bar(chunk.frames() as u64);
                    
                    let progress_file = Path::new(&self.project.args.temp).join("done.json");
```

Maybe alternative pathways could be implemented for x26* encoders or they can disable this feature for now.

GreatValueCreamSoda:
X265 does not output raw .ivf files. It outputs raw HEVC streams.

av1an-core/src/broker.rs
```diff
@@ -258,8 +294,6 @@ impl Broker<'_> {
                         );
                         return Err(Some(e));
                     }
-                    // avoids double-print of the error message as both a WARN and ERROR,
```
BoatsMcGee, line 261: Is there any reason why comments like these need to be removed? This one in particular is quite important because the reason why it's printed here is not obvious unless you also know how the function called handles this error case.

In my opinion, Av1an codebase is in need of more comments rather than fewer. The "space padding" and "CTRL+C" comments could be considered obvious based on their surrounding code but that doesn't apply for this one.

av1an-core/src/vmaf.rs
```diff
@@ -284,6 +285,234 @@ pub fn run_vmaf(
...
+            source_pipe_stderr: String::new().into(),
+            ffmpeg_pipe_stderr: None,
+            stderr:             "Y, U, V VMAF files were not created".to_string().into(),
+            stdout:             String::new().into(),
+        }));
+    }
+
+    let y_scores = read_vmaf_file(&vmaf_y_path).map_err(|e| {
+        let dummy_output =
+            Command::new("false").output().unwrap_or_else(|_| std::process::Output {
```
BoatsMcGee, line 420: Running a command called "false" could lead to unintended consequences should the user's environment have a command or script called "false".

Is this workaround necessary for getting the ExitStatus? If so, could we instead invoke a command that we know will behave as expected, such as "ls" or "av1an" itself?

Additionally, could we reduce the repetition of this code to a single "dummy_output" that is declared before it is used repeatedly? Perhaps a single function that replaces the "YUV" value in the STDERR string?

av1an-core/src/vmaf.rs
```diff
@@ -292,10 +521,6 @@ pub fn read_vmaf_file(file: impl AsRef<Path>) -> Result<Vec<f64>, serde_json::Er
     Ok(v)
 }
 
-/// Read a certain percentile VMAF score from the VMAF json file
```
BoatsMcGee, line 295: I see no reason why this comment needs to be removed - it even has a warning and information that would be helpful for those unfamiliar with the code or the language.

av1an-core/src/vmaf.rs
```diff
@@ -305,21 +530,49 @@ pub fn read_weighted_vmaf<P: AsRef<Path>>(
...
+        };
+        let index = index.min(scores.len() - 1);
+
+        Ok(scores[index])
+    }
+
+    inner(file.as_ref(), percentile)
+}
+
+pub fn read_weighted_vmaf_alt<P: AsRef<Path>>(
```
BoatsMcGee, line 550: This function seems identical to the non-alt version which also seems to be needlessly modified despite this new function being introduced.

I think a better approach would be to modify the existing function to suit the needs of the alt by passing framerate as an Option. On second thought, framerate appears to be unused and this function appears to be needlessly duplicated.

av1an/src/main.rs
```diff
@@ -660,6 +660,16 @@ pub struct CliOpts {
     /// If not specified, the default value is used (chosen per encoder).
     #[clap(long, help_heading = "Target Quality")]
     pub max_q: Option<u32>,
+
+    /// Use alternative perceptually weighted VMAF with motion compensation
+    /// disabled
+    #[clap(long, help_heading = "Target Quality")]
+    pub alt_vmaf: bool,
```
BoatsMcGee, line 667: This is arguably a nitpick but this option name does not convey its purpose in entirety and could potentially cause conflict should we want to add more alternatives to VMAF in the future. The most important part of this feature is that the VMAF used disables motion compensation so a more descriptive name would be "motionless_vmaf" or "still_vmaf" but those aren't great either. Neither would "weighted_vmaf".

Additionally, VMAF is used outside of Target Quality in plotting. This alternative does not appear to be implemented for that and users could be confused as to why it doesn't despite the name.

I suggest an alternative approach that avoids future conflicts with potential additions or changes to VMAF, offering more flexibility to the user while remaining descriptive. Instead of a boolean, use an enum to enable or disable features. For example, this could be `--probing-vmaf-features [default?, weighted, motionless]`. That way, users can disable motion compensation without using the weighted formula. While it hasn't been used in this codebase yet, it's possible to allow users to specify an option multiple times to "select" multiple options without spamming boolean flags for each item. If you need more details, I can dig up the clap documentation. The latter is definitely an option that also avoids this naming and technical debt issue. I think both should be considered.

av1an/src/main.rs
```diff
@@ -660,6 +660,16 @@ pub struct CliOpts {
...
+
+    /// Percentile to use for quality targeting (0.01 = 1st percentile, 0.10 =
+    /// 10th percentile)
+    #[clap(long, default_value_t = 0.01, help_heading = "Target Quality")]
+    pub percentile: f64,
```
BoatsMcGee, line 672: This option does not follow the convention already set by other similar options. Options that change probing behavior are prefixed with "probing-" which helps the user understand better what the option is intended for.

This option is also not as flexible for the user as it can be. I will preface that I agree the percentile choice is not optimal and if we are to present the user with an option for an alternative, it should be an option that provides the most value. We are already paying the cost of increased complexity in the user experience by presenting an option, we might as well make the most of it.

### Alternative 1

Provide multiple choices of what statistic the user would like to use to map-reduce the frame scores in probes. This can be done in several ways but the quickest example I can provide is in [#BoatsMcGee/Av1an](https://github.com/BoatsMcGee/Av1an/commit/03af611c67fcf86d39464dc8cbeb2aac068b3e1a#diff-3f84f500d60d8de21b14cba4d30b00526da8ba7a191d371f4bd1bf68f433ebecR537). This adds a `--probe-stat` option to give the user more choice in their preferred method.  I would rename it to `--probing-stat(istic)` for consistency. You have my explicit permission to borrow and improve from this.

### Alternative 2 (Recommended)
We should keep it simple for the users. I would surmise most users of Target Quality use it for the convenience and simplicity it offers rather than for the niche that this option provides. The aforementioned branch was used in an experiment to determine the correlation between quantizer prediction accuracy, the statistic used for probing, and the speed at which the probe was encoded. The data suggests that the optimal statistic used depends on the speed at which the probe was encoded. The speed at which the probe was encoded had the most significant impact so if users need more accuracy, they should primarily use a slower speed and optionally `--probe-slow`. If their final output has inconsistencies, they can use `--max-q` and a slower encoding speed/preset for more consistent results. However, if we use the probe speed specified, we can choose the optimal statistic to further improve the prediction accuracy and keep the options simple for the user. Briefly, another feature would be to improve probe accuracy by probing at different speeds based on the quantizer. This would be a simple addition: add "auto" to the `--probing-speed` enum, which would signal Target Quality to probe with a preset/speed based on the quantizer.

I highly recommend running that experiment with the changes from this PR to gather the data and insight necessary to make a judgement based on data.

av1an-core/src/broker.rs
```diff
@@ -45,7 +51,7 @@ impl Debug for StringOrBytes {
         match self {
             Self::String(s) => {
                 if f.alternate() {
-                    f.write_str(&textwrap::indent(s, /* 8 spaces */ "        "))?;
```
BoatsMcGee, line 48: This comment appears extraneous and could be improved but it suggests that there are 8 spaces for a good reason - in order to format text on the terminal in relation to other text. I don't think removing it is an improvement and it also isn't related to this PR.

av1an-core/src/target_quality.rs
```diff
@@ -1,31 +1,18 @@
-#[cfg(test)]
-mod tests;
-
-use std::{
-    cmp,
-    cmp::Ordering,
-    convert::TryInto,
-    fmt::Error,
-    path::{Path, PathBuf},
-    process::Stdio,
```
BoatsMcGee, line 10: Is there any reason why Stdio needs to be removed and subsequently fully qualified every time it is used later in this file?

av1an-core/src/target_quality.rs
```diff
@@ -46,20 +33,17 @@ pub struct TargetQuality {
...
         worker_id: Option<usize>,
-    ) -> Result<u32, Box<EncoderCrash>> {
-        let mut vmaf_cq = vec![];
-        let frames = chunk.frames();
-
-        // Make middle probe
-        let middle_point = (self.min_q + self.max_q) / 2;
-        let last_q = middle_point;
+    ) -> anyhow::Result<u32> {
+        let mut history: Vec<(u32, f64, PathBuf)> = vec![];
```
BoatsMcGee, line 46: If this were a struct with named properties rather than a tuple, "history" would be an acceptable name. However, I renamed this variable in PR #988 and I believe that name is a better fit and could still be improved upon. It also describes what the items in the tuple are without having to check this line when working on code below.

PathBuf is also added here and is never used. Why does the VMAF file needs to be in the score_quantizer vector? Although PR #988 has not yet been evaluated or merged, it would cause a direct conflict that cannot be resolved/refactored without removing it.

av1an-core/src/target_quality.rs
```diff
@@ -251,33 +216,50 @@ impl TargetQuality {
...
-            self.vmaf_threads,
-        )?;
+        let probe_name = std::path::Path::new(&chunk.temp)
+            .join("split")
+            .join(format!("v_{q}_{}.ivf", chunk.index));
+        let fl_path = std::path::Path::new(&chunk.temp)
+            .join("split")
+            .join(format!("{}.json", chunk.index));
+
+        if self.alt_vmaf {
```
BoatsMcGee, line 235: As both functions return the same result and do very similar things, it would be better to consolidate the change into the preexisting function as I mentioned in the review for the implementation. Essentially, DRY.

av1an-core/src/target_quality.rs
```diff
@@ -287,35 +269,62 @@ impl TargetQuality {
         &self,
         chunk: &mut Chunk,
         worker_id: Option<usize>,
-    ) -> Result<(), Box<EncoderCrash>> {
+    ) -> anyhow::Result<()> {
```
BoatsMcGee, line 272: Why has the error been removed? It could and should be handled by the caller - even if it currently isn't.

av1an-core/src/target_quality.rs
```diff
@@ -287,35 +269,62 @@ impl TargetQuality {
...
         chunk.tq_cq = Some(self.per_shot_target_quality(chunk, worker_id)?);
         Ok(())
     }
 }
 
-pub fn weighted_search(num1: f64, vmaf1: f64, num2: f64, vmaf2: f64, target: f64) -> usize {
-    let dif1 = (transform_vmaf(target) - transform_vmaf(vmaf2)).abs();
-    let dif2 = (transform_vmaf(target) - transform_vmaf(vmaf1)).abs();
+fn predict_crf(low: u32, high: u32, history: &[(u32, f64, PathBuf)], target: f64) -> u32 {
```
BoatsMcGee, line 278: Pathbuf appears to never be used in this function and this function requires the tuple item of this type. Why was this added? Do we need it?

av1an-core/src/target_quality.rs
```diff
@@ -287,35 +269,62 @@ impl TargetQuality {
...
-    num1.mul_add(dif1 / tot, num2 * (dif2 / tot)).round() as usize
-}
+    if crf_score_map.len() >= 3 {
+        let (scores, crfs): (Vec<f64>, Vec<f64>) =
+            crf_score_map.iter().map(|(crf, score)| (*score, *crf as f64)).unzip();
+
+        let keys: Vec<Key<f64, f64>> = scores
+            .iter()
+            .zip(crfs.iter())
+            .map(|(score, crf)| Key::new(*score, *crf, Interpolation::CatmullRom))
```
BoatsMcGee, line 293: `interpolate_target_q` previously used Linear - likely as a default or standard. This reimplementation of that function in here uses a different Interpolation method but would this curve be appropriate for every encoder and metric? Some metrics are not similar to VMAF, such as XPSNR or butteraugli.

Additionally, `lagrange_bisect` was removed but based on my limited understanding of lagrange interpolation,  that appears to be a much better algorithm for predicting quantizers based on 3 or more data points. It's also a more generalized approach that does not assume the "shape" of the curve or degree of the polynomial like the previous `interpolate_target_q` and this change to `CatmullRom`.

I highly recommend evaluating `lagrange_bisect` and using that instead. I have briefly tested it locally and found the results satisfactory and at the very least in working order. As for why it was never implemented after 3 years of remaining dormant I can only surmise that it was forgotten. I think this is a good opportunity to implement it.

av1an-core/src/target_quality.rs
```diff
@@ -287,35 +269,62 @@ impl TargetQuality {
...
+
+fn within_tolerance(score: f64, target: f64) -> bool {
+    (score - target).abs() / target < 0.01
 }
 
-/// Returns auto detected amount of threads used for vmaf calculation
 pub fn vmaf_auto_threads(workers: usize) -> usize {
     const OVER_PROVISION_FACTOR: f64 = 1.25;
 
-    // Logical CPUs
```
BoatsMcGee, line 318: These comments clarify things and should not be removed.

av1an-core/src/target_quality.rs
```diff
@@ -326,40 +335,8 @@ pub fn vmaf_auto_threads(workers: usize) -> usize {
     )
 }
```
BoatsMcGee, line 337: Although the implementation of this function has been modified and hoisted elsewhere, would it not have been better to keep this function and modify it to suit the new change instead?

av1an-core/src/target_quality.rs
```diff
@@ -402,69 +376,3 @@ pub const fn adapt_probing_rate(rate: usize) -> usize {
...
-                    den *= i64::from(*x0) - i64::from(*x1);
-                }
-            }
-            y0 * num as f64 / den as f64
-        })
-        .sum()
-}
-
-#[allow(unused)]
-fn lagrange_bisect(p: &[(u32, f64)], y: f64) -> (u32, f64) {
```
BoatsMcGee, line 432: Although this code has remained dormant, it is a more generalized approach to predicting a value on a polynomial without assuming the degree of the polynomial. Instead of removing it, I highly recommend evaluating if we could use this already existing and implemented code to perform the same task.

av1an-core/src/vmaf.rs
```diff
@@ -284,6 +285,234 @@ pub fn run_vmaf(
...
+            });
+        Box::new(EncoderCrash {
+            exit_status:        dummy_output.status,
+            source_pipe_stderr: String::new().into(),
+            ffmpeg_pipe_stderr: None,
+            stderr:             format!("Failed to serialize VMAF results: {}", e).into(),
+            stdout:             String::new().into(),
+        })
+    })?;
+    fs::write(stat_file, json_str).map_err(|e| {
```
BoatsMcGee, line 497: Instead of rewriting the file with JSON and adding the "fs" dependency in the calculation function, why not use the existing parser function to read, evaluate, and return the value?

av1an-core/Cargo.toml
```diff
@@ -103,3 +103,4 @@ vapoursynth_new_api = [
 [lints.clippy]
 inline_always = "warn"
 missing_inline_in_public_items = "warn"
+uninlined_format_args = "allow"
```
BoatsMcGee, line 106: Why was this rule added? I don't see the relationship between this change and the intent of the PR.

av1an-core/src/target_quality.rs
```diff
@@ -70,101 +54,82 @@ impl TargetQuality {
...
             }
         }
 
-        let (q, q_vmaf) = interpolated_target_q(vmaf_cq.clone(), self.target);
+        let good_results: Vec<&(u32, f64, PathBuf)> = history
+            .iter()
+            .filter(|(_, score, _)| within_tolerance(*score, self.target))
+            .collect();
+
+        debug!(
```
BoatsMcGee, line 98: These logs should be handled by `log_probes` instead to centralize and standardize the logging.

av1an-core/src/target_quality.rs
```diff
@@ -239,8 +206,6 @@ impl TargetQuality {
             };
 
             let source_pipe_output = source_pipe.wait_with_output().await.unwrap();
-
-            // TODO: Expand EncoderCrash to handle io errors as well
```
BoatsMcGee, line 243: Although this TODO should never have made it into "production" by strict standards, it is still a task that needs to be done as this comment suggests. We should post a new Issue for this before we ever remove this or any comment like it.

BoatsMcGee (changes requested):
I approve of the addition to exit the probing loop when a certain tolerance threshold is reached - I think it's a great approach. Simplifying the `per_shot_target_quality` function is also a refreshing improvement in readability though it could be further improved.

However I have several concerns across these review items. A common issue is that there are extraneous changes to the code that are not related to the PR. There is also a lot of needless repetition in some functions as well as some functions themselves.

Modifying a CLI and its "API", especially one that has been around for years, should be done with serious consideration to the average user experience, users exploring the "help" output, and users who wish to take advantage of the new features. One of the reviews details the last point further. Additionally, we should avoid adding technical debt whenever we can and there are many changes that don't consider the possibility of expansion or new features.

That last point is immediately evident in that there are some changes that are incompatible with the changes in PR #988 and more details are in some of the reviews.

Thanks
\- Boats M.

av1an-core/Cargo.toml
emrakyz: Soichiro probably added this after my PR.

I think this syntax can create a warning (and similar ones):
`format!("v_{}_{}.ivf", optimal_q, chunk.index)`

Rather than updating all these format strings to use the newer inline syntax throughout the codebase, Soi chose to disable this warning.

emrakyz:
Hey @BoatsMcGee 

Thanks for all of your reviews and insights.

I'll adress all of them & fix things along the way according to the reviews and the small error we get because of the `.ivf` extension.

This PR became a mess after Soichiro mistakenly broke it :)

So, I'll open a new PR to be able to continue.

shssoichiro:
Closing for #1014

PR #1014, opened 2025-06-11
New Probing Options & Better Search & Cleanups

### LINKED ISSUE #818: --probe-slow duplicate coding causes the whole process to increase in length
--probe-slow when enabled will encode according to the given parameters, then there is no need to encode again when the target crf is computed, as it is already encoded. The file generated in the split folder and the file generated by specifying the crf in the encode file are the same file. Therefore, for target quality mode when --probe-slow is enabled, you only need to copy the files in the split folder that correspond to the calculated crf value to the encode folder. This will reduce the time needed to encode the entire video at once. In a 4+ minute video I tested, it was about 30% less time. (Of course this also depends on the number of --probes)

emrakyz:
Fixes #818, #1008, #1009

**EDIT after reviews:**
- Fixed some problems + some reviews from Boats
- Modernized variables (inlined named variables)
- While I was at it, I fixed some easy TODO item (ilog2). It was waiting for a stability from a decade ago :)
- Fixed encoder specific intermediate file extension stuff.
- Removed redundancies and tried to streamline further. 

This PR attempts to add **3 new CLI flags** and improve/change the CRF search algorithm by implementing Catmull-Rom Spline Interpolation, effectively making CRF search more accurate and faster and preventing its unwanted early-exit behavior.

```
 --probing-percent 0.10 --probing-vmaf-features default weighted motionless --target-quality 90 --probes 99 --probing-rate 1 --probing-speed veryslow --probe-slow --min-q 1 --max-q 70
```

It also adds alternative VMAF methods which disables the motion compensation that artifically boosts scores and perceptual YUV weighthing (`4 + 1 + 1 / 6`). 

- Start the search loop

# Phase 1 (No History)

# Phase 2 (two data points)

# Phase 3: Three+ Data

# Conditions

# Termination

# Selection

# Extra Probing Features

# Added CLI Flags
image

image


emrakyz:
There was a small git problem. I'll quickly send a commit now.

EDIT: It was just the clap import.

av1an-core/src/target_quality.rs
```diff
@@ -191,46 +190,48 @@ impl TargetQuality {
...
-            let source_pipe_stdout: Stdio = source.stdout.take().unwrap().try_into().unwrap();
+            let source_pipe_stdout: std::process::Stdio =
+                source.stdout.take().unwrap().try_into().unwrap();
 
             let mut source_pipe = if let [ffmpeg, args @ ..] = &*cmd.0 {
                 tokio::process::Command::new(ffmpeg)
                     .args(args)
                     .stdin(source_pipe_stdout)
-                    .stdout(Stdio::piped())
+                    .stdout(std::process::Stdio::piped())
```
shssoichiro, line 211: Was there an issue that led to needing to use std's Stdio instead of tokio's here?

emrakyz: Yes. But I couldn't remember.

I guess if you use `tokio::process::Stdio::piped()`, you get a `tokio::process::ChildStdout` when you call `.stdout.take()`. But then when you try to convert that to `std::process::Stdio` for the next process in the chain (using `.try_into())`, I guess you might run into conversion issues or the conversion might not be available.

By using `std::process::Stdio::piped()` directly, the stdout handle is already in the std ecosystem. Makes the subsequent conversion smoother when chaining to the next process.

shssoichiro (approved):

PR #1016, opened 2025-06-12
Refactor TQ & Consolidate Stats (Cross-PR with Boats)

emrakyz:
- Some additions for available statistical models for the target quality decisions.
- Refactored TQ & Some clean-ups and enhanced logging.
- Better error handling.

This PR aims to establish a foundation for #988, which was developed through discussions with @BoatsMcGee, primarily regarding their modifications.

### DIFF
--- a/av1an-core/src/concat.rs
+++ b/av1an-core/src/concat.rs
@@ -162,7 +162,7 @@ pub fn mkvmerge(
     output: &Path,
     encoder: Encoder,
     num_chunks: usize,
-    output_fps: Rational64,
+    output_fps: Option<Rational64>,
 ) -> anyhow::Result<()> {
     // mkvmerge does not accept UNC paths on Windows
     #[cfg(windows)]
@@ -242,20 +242,24 @@ pub fn mkvmerge_options_json(
     encoder: Encoder,
     output: &str,
     audio: Option<&str>,
-    output_fps: Rational64,
+    output_fps: Option<Rational64>,
 ) -> String {
     let mut file_string = String::with_capacity(64 + 12 * num);
     write!(file_string, "[\"-o\", {output:?}").unwrap();
     if let Some(audio) = audio {
         write!(file_string, ", {audio:?}").unwrap();
     }
-    write!(
-        file_string,
-        ", \"--default-duration\", \"0:{}/{}fps\", \"[\"",
-        output_fps.numer(),
-        output_fps.denom()
-    )
-    .unwrap();
+    if let Some(output_fps) = output_fps {
+        write!(
+            file_string,
+            ", \"--default-duration\", \"0:{}/{}fps\", \"[\"",
+            output_fps.numer(),
+            output_fps.denom()
+        )
+        .unwrap();
+    } else {
+        file_string.push_str(", \"[\"");
+    }
     for i in 0..num {
         write!(
             file_string,
--- a/av1an-core/src/concat/tests.rs
+++ b/av1an-core/src/concat/tests.rs
@@ -4,7 +4,13 @@ use super::*;
 
 #[test]
 fn test_mkvmerge_options_json_no_audio() {
-    let result = mkvmerge_options_json(2, Encoder::aom, "output.mkv", None, Rational64::new(30, 1));
+    let result = mkvmerge_options_json(
+        2,
+        Encoder::aom,
+        "output.mkv",
+        None,
+        Some(Rational64::new(30, 1)),
+    );
     assert_eq!(
         result,
         r#"["-o", "output.mkv", "--default-duration", "0:30/1fps", "[", "00000.ivf", "00001.ivf","]"]"#
@@ -18,7 +24,7 @@ fn test_mkvmerge_options_json_with_audio() {
         Encoder::aom,
         "output.mkv",
         Some("audio.mkv"),
-        Rational64::new(30, 1),
+        Some(Rational64::new(30, 1)),
     );
     assert_eq!(
         result,
--- a/av1an-core/src/context.rs
+++ b/av1an-core/src/context.rs
@@ -391,7 +391,19 @@ impl Av1anContext {
                         self.args.output_file.as_ref(),
                         self.args.encoder,
                         total_chunks,
-                        fps_ratio,
+                        if self.args.ignore_frame_mismatch {
+                            info!(
+                                "`--ignore-frame-mismatch` set. Don't force output FPS, as an FPS \
+                                 changing filter might have been applied."
+                            );
+                            None
+                        } else {
+                            debug!(
+                                "`--ignore-frame-mismatch` not set. Forcing output FPS to \
+                                 {fps_ratio} with mkvmerge."
+                            );
+                            Some(fps_ratio)
+                        },
                     )?;
                 },
                 ConcatMethod::FFmpeg => {
--- a/av1an/src/main.rs
+++ b/av1an/src/main.rs
@@ -621,9 +621,14 @@ pub struct CliOpts {
     #[clap(long, default_value_t = 4, help_heading = "Target Quality")]
     pub probes: u32,
 
-    /// Framerate for probes, 1 - original
-    #[clap(long, default_value_t = 1, help_heading = "Target Quality")]
-    pub probing_rate: u32,
+    /// Only use every nth frame for VMAF calculation, while probing.
+    ///
+    /// WARNING: The resulting VMAF score might differ from if all the frames
+    /// were used; usually it should be lower, which means to get the same
+    /// quality you must also usually lower the --target-quality. Going higher
+    /// than n=4 usually results in unusable scores, so this is disabled.
+    #[clap(long, default_value_t = 1, value_parser = clap::value_parser!(u16).range(1..=4), help_heading = "Target Quality")]
+    pub probing_rate: u16,
 
     /// Speed for probes, defaults to veryfast
     ///
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,8 @@
+﻿coverage:
+  status:
+    project:
+      default:
+        informational: true # Makes status checks informational only
+    patch:
+      default:
+        informational: true # Makes patch coverage informational too


master-of-zen (commented):
LGTM 👍🏽🥇

av1an-core/src/target_quality.rs
```diff
@@ -79,91 +84,123 @@ impl TargetQuality {
...
+                .iter()
+                .any(|(quantizer, _)| *quantizer == next_quantizer)
+            {
+                // Predicted quantizer has already been probed
+                let (last_quantizer, last_score) = quantizer_score_history
+                    .iter()
+                    .find(|(quantizer, _)| *quantizer == next_quantizer)
+                    .unwrap();
+                log_probes(
+                    &mut quantizer_score_history.clone(),
```
shssoichiro: Is this working as expected? If we intend for `log_probes` to write back to the `quantizer_score_history`, the clone here will make it so that doesn't happen. If we need to clone to avoid the borrow checker, we should be cloning in a non-mutable usage site instead.

BoatsMcGee: `log_probes` sorts the vector before logging it. The vector is no longer used after that. I believe it was mutable because it avoided making an unnecessary copy in memory. Due to the above iterator this particular call seems to be problematic.

I've changed the `log_probes` as you suggested. The vector is borrowed then copied in the function (if I understand that correctly) with `to_vec()` then subsequently sorted and used in the log.

As another simplification to the logic I changed `final_quantizer_score` to use `if let Some()` which avoids iterating over the array twice for single code-path - I'm starting to get used to these Rust idioms more now. I also changed the `debug!` macros used in `predict_quantizer` to `trace!` instead as it's relatively verbose and non-important given how we're using `debug!`.

<details>
<summary>Here is a diff containing these changes:</summary>

```diff
diff --git a/av1an-core/src/target_quality.rs b/av1an-core/src/target_quality.rs
index 0f8a8fd..bb98079 100644
--- a/av1an-core/src/target_quality.rs
+++ b/av1an-core/src/target_quality.rs
@@ -11,7 +11,7 @@ use clap::ValueEnum;
 use ffmpeg::format::Pixel;
 use serde::{Deserialize, Serialize};
 use splines::{Interpolation, Key, Spline};
-use tracing::debug;
+use tracing::{debug, trace};
 
 use crate::{
     broker::EncoderCrash,
@@ -96,24 +96,20 @@ impl TargetQuality {
                 self.target,
             );
 
-            if quantizer_score_history
+            if let Some((quantizer, score)) = quantizer_score_history
                 .iter()
-                .any(|(quantizer, _)| *quantizer == next_quantizer)
+                .find(|(quantizer, _)| *quantizer == next_quantizer)
             {
                 // Predicted quantizer has already been probed
-                let (last_quantizer, last_score) = quantizer_score_history
-                    .iter()
-                    .find(|(quantizer, _)| *quantizer == next_quantizer)
-                    .unwrap();
                 log_probes(
-                    &mut quantizer_score_history.clone(),
+                    &quantizer_score_history,
                     self.target,
                     chunk.frames() as u32,
                     self.probing_rate as u32,
                     self.probing_speed,
                     &chunk.name(),
-                    *last_quantizer,
-                    *last_score,
+                    *quantizer,
+                    *score,
                     SkipProbingReason::None,
                 );
                 break;
@@ -138,7 +134,7 @@ impl TargetQuality {
 
             if score_within_tolerance || quantizer_score_history.len() >= self.probes as usize {
                 log_probes(
-                    &mut quantizer_score_history,
+                    &quantizer_score_history,
                     self.target,
                     chunk.frames() as u32,
                     self.probing_rate as u32,
@@ -164,7 +160,7 @@ impl TargetQuality {
             // Ensure quantizer limits are valid
             if lower_quantizer_limit > upper_quantizer_limit {
                 log_probes(
-                    &mut quantizer_score_history,
+                    &quantizer_score_history,
                     self.target,
                     chunk.frames() as u32,
                     self.probing_rate as u32,
@@ -182,14 +178,16 @@ impl TargetQuality {
             }
         }
 
-        let history_within_tolerance: Vec<&(u32, f64)> = quantizer_score_history
-            .iter()
-            .filter(|(_, score)| within_tolerance(*score, self.target))
-            .collect();
-
-        let final_quantizer_score = if !history_within_tolerance.is_empty() {
-            history_within_tolerance.iter().max_by_key(|(quantizer, _)| *quantizer).unwrap()
+        let final_quantizer_score = if let Some(highest_quantizer_score_within_tolerance) =
+            quantizer_score_history
+                .iter()
+                .filter(|(_, score)| within_tolerance(*score, self.target))
+                .max_by_key(|(quantizer, _)| *quantizer)
+        {
+            // Multiple probes within tolerance, choose the highest
+            highest_quantizer_score_within_tolerance
         } else {
+            // No quantizers within tolerance, choose the quantizer closest to target
             quantizer_score_history
                 .iter()
                 .min_by(|(_, score1), (_, score2)| {
@@ -420,14 +418,14 @@ fn predict_quantizer(
     let spline = Spline::from_vec(keys);
     let predicted_quantizer = spline.sample(target).unwrap_or_else(|| {
         // Probes do not fit Catmull-Rom curve, fallback to Linear
-        debug!("Probes do not fit Catmull-Rom curve, falling back to Linear");
+        trace!("Probes do not fit Catmull-Rom curve, falling back to Linear");
         let keys = sorted_quantizer_score_history
             .iter()
             .map(|(quantizer, score)| Key::new(*score, *quantizer as f64, Interpolation::Linear))
             .collect();
         Spline::from_vec(keys).sample(target).unwrap_or_else(|| {
             // Probes do not fit Catmull-Rom curve or Linear, fallback to binary search
-            debug!("Probes do not fit Linear curve, falling back to binary search");
+            trace!("Probes do not fit Linear curve, falling back to binary search");
             binary_search as f64
         })
     });
@@ -464,7 +462,7 @@ pub enum SkipProbingReason {
 
 #[allow(clippy::too_many_arguments)]
 pub fn log_probes(
-    quantizer_score_history: &mut [(u32, f64)],
+    quantizer_score_history: &[(u32, f64)],
     target: f64,
     frames: u32,
     probing_rate: u32,
@@ -475,7 +473,8 @@ pub fn log_probes(
     skip: SkipProbingReason,
 ) {
     // Sort history by quantizer
-    quantizer_score_history.sort_by_key(|(quantizer, _)| *quantizer);
+    let mut sorted_quantizer_scores = quantizer_score_history.to_vec();
+    sorted_quantizer_scores.sort_by_key(|(quantizer, _)| *quantizer);
 
     debug!(
         "chunk {name}: Target={target}, P-Rate={rate}, P-Speed={speed:?}, {frame_count} frames
@@ -486,7 +485,7 @@ pub fn log_probes(
         rate = probing_rate,
         speed = ProbingSpeed::from_repr(probing_speed.unwrap_or(4) as usize),
         frame_count = frames,
-        history = quantizer_score_history,
+        history = sorted_quantizer_scores,
         suffix = match skip {
             SkipProbingReason::None => "",
             SkipProbingReason::QuantizerTooHigh => "Early Skip High Quantizer",

```
</details>

Thanks,
\- Boats M.

BoatsMcGee:
These warning mentioning other metrics do not belong here (yet). We don't want to confuse users with metrics that aren't available in the current program. Don't worry, I'll add them back and improve on the language when I merge with #988. "can only be used" should be changed to "should only be used with ..." as we don't want to suggest that we don't allow it - we just don't recommend it. The language can be improved further but we'll leave it for the related PR. I've included it here below as a reminder.

```rs
/// Warning:
    ///   'root-mean-square, Root Mean Square' should only be used with reverse metrics such as 'BUTTERAUGLI'.
    ///   'HARMONIC' only works when there is no negative score. Use with caution on 'SSIMU2'.
```

"rms" has been changed to "root-mean-square" for consistency and clarity (not that anyone would confuse it for Royal Mail Ship). The validation and parsing match was missing "harmonic" and "root-mean-square". The documentation was also missing "root-mean-square". I also fixed the issue in the documentation that caused Copilot to be confused about the option's name. It should be `--probing-stat`, not `--probing-statistic`. That's my mistake as I wrote the documentation while considering changing the option name to be more verbose - but I don't think it improves clarity so "stat" is sufficient.

<details>
<summary>Here is the diff containing all these changes:</summary>

```diff
diff --git a/av1an-core/src/lib.rs b/av1an-core/src/lib.rs
index 3839958..5198b68 100644
--- a/av1an-core/src/lib.rs
+++ b/av1an-core/src/lib.rs
@@ -534,7 +534,7 @@ pub enum ProbingStatisticName {
     Minimum = 6,
     #[strum(serialize = "maximum")]
     Maximum = 7,
-    #[strum(serialize = "rms")]
+    #[strum(serialize = "root-mean-square")]
     RootMeanSquare = 8,
 }
 
diff --git a/av1an/src/main.rs b/av1an/src/main.rs
index e49329d..2b5fef7 100644
--- a/av1an/src/main.rs
+++ b/av1an/src/main.rs
@@ -695,11 +695,11 @@ pub struct CliOpts {
     ///   mode                       - Most common integer-rounded value
     ///   minimum                    - Lowest value
     ///   maximum                    - Highest value
-    ///   rms                        - Root Mean Square (quadratic mean)
+    ///   root-mean-square           - Root Mean Square (quadratic mean)
     ///
     /// Warning:
-    ///   'RMS, Root Mean Square' can only be used with reverse metrics such as 'BUTTERAUGLI'.
-    ///   'HARMONIC' only works when there is no negative score. Use with caution on 'SSIMU2'.
+    ///   root-mean-square, Root Mean Square' should only be used with reverse metrics.
+    ///   'HARMONIC' works as expected when there are no negative scores. Use with caution.
     #[clap(long, default_value_t = String::from("percentile=1"), help_heading = "Target Quality", verbatim_doc_comment)]
     pub probing_stat: String,
 }
@@ -723,6 +723,14 @@ impl CliOpts {
                         name:  ProbingStatisticName::Mean,
                         value: None,
                     },
+                    "harmonic" => ProbingStatistic {
+                        name:  ProbingStatisticName::Harmonic,
+                        value: None,
+                    },
+                    "root-mean-square" => ProbingStatistic {
+                        name:  ProbingStatisticName::RootMeanSquare,
+                        value: None,
+                    },
                     "median" => ProbingStatistic {
                         name:  ProbingStatisticName::Median,
                         value: None,
diff --git a/site/src/Cli/target_quality.md b/site/src/Cli/target_quality.md
index 154d126..37cd974 100644
--- a/site/src/Cli/target_quality.md
+++ b/site/src/Cli/target_quality.md
@@ -6,7 +6,7 @@ Name | Flag | Type | Default
 [Probes](#probes---probes) | `--probes` | Integer | `4`
 [Probing Rate](#probing-rate---probing-rate) | `--probing-rate` | Integer | `1`
 [Probing Speed](#probing-speed---probing-speed) | `--probing-speed` | `PROBING_SPEED` |
-[Probing Statistic](#probing-statistic---probing-statistic) | `--probing-statistic` | String | `percentile=1`
+[Probing Statistic](#probing-statistic---probing-stat) | `--probing-stat` | String | `percentile=1`
 [Probe Slow](#probe-slow---probe-slow) | `--probe-slow` || 
 [Minimum Quantizer](#minimum-quantizer---min-q) | `--min-q` | Integer | Based on Encoder
 [Maximum Quantizer](#maximum-quantizer---max-q) | `--max-q` | Integer | Based on Encoder
@@ -73,7 +73,7 @@ Can be any of the following:
 
 If not specified, `veryfast` is used unless `--probe-slow` is specified.
 
-## Probing Statistic `--probing-statistic`
+## Probing Statistic `--probing-stat`
 
 Statistical method for calculating target quality from sorted probe results.
 
@@ -84,6 +84,7 @@ Can be any of the following:
 * `mean` - Arithmetic mean (average)
 * `median` - Middle value
 * `harmonic` - Harmonic mean (emphasizes lower scores)
+* `root-mean-square` - Root mean square (quadratic mean)
 * `percentile=<FLOAT>` - Percentile of a specified `<FLOAT>` value, where `<FLOAT>` is a value between 0.0 and 100.0
 * `standard-deviation=<FLOAT>` - Standard deviation distance from mean (σ) clamped by the minimum and maximum probe scores of a specified `<FLOAT>` value, where `<FLOAT>` can be a positive or negative value
 * `mode` - Most common integer-rounded value
diff --git a/site/src/av1an.md b/site/src/av1an.md
index e3dda23..bfbf75b 100644
--- a/site/src/av1an.md
+++ b/site/src/av1an.md
@@ -89,7 +89,7 @@ Name | Flag | Type | Default
 [Probes](./Cli/target_quality.md#probes---probes) | `--probes` | Integer | `4`
 [Probing Rate](./Cli/target_quality.md#probing-rate---probing-rate) | `--probing-rate` | Integer | `1`
 [Probing Speed](./Cli/target_quality.md#probing-speed---probing-speed) | `--probing-speed` | `PROBING_SPEED` |
-[Probing Statistic](./Cli/target_quality.md#probing-statistic---probing-statistic) | `--probing-statistic` | String | `percentile=1`
+[Probing Statistic](./Cli/target_quality.md#probing-statistic---probing-stat) | `--probing-stat` | String | `percentile=1`
 [Probe Slow](./Cli/target_quality.md#probe-slow---probe-slow) | `--probe-slow` || 
 [Minimum Quantizer](./Cli/target_quality.md#minimum-quantizer---min-q) | `--min-q` | Integer | Based on Encoder
 [Maximum Quantizer](./Cli/target_quality.md#maximum-quantizer---max-q) | `--max-q` | Integer | Based on Encoder
\ No newline at end of file

```

</details>

emrakyz:
@BoatsMcGee 

Thanks Boats! All resolved.

av1an-core/src/target_quality.rs
```diff
@@ -79,91 +77,121 @@ impl TargetQuality {
...
         } else {
-            history
+            // No quantizers within tolerance, choose the quantizer closest to target
+            quantizer_score_history
                 .iter()
-                .min_by(|(_, s1, _), (_, s2, _)| {
-                    let d1 = (s1 - self.target).abs();
-                    let d2 = (s2 - self.target).abs();
-                    d1.partial_cmp(&d2).unwrap_or(Ordering::Equal)
+                .min_by(|(_, score1), (_, score2)| {
```
shssoichiro, line 186: Nit: Using `min_by_key` would probably be simpler here, but I don't want to hold up the PR on that

PR #1017, opened 2025-06-13
fix fbba019e: forcing output FPS breaks using FPS changing filter

t-nil:
Hi,

I use FFmpeg on an input filter to change fps of some of my clips: `--ignore-frame-mismatch --ffmpeg "-vf scale=-1:\\'min(ih,1080)\\':flags=lanczos,fps=15"`. After updating to newest master I found that fbba019e broke my method: since now, the output muxer is forced unconditionally to emit videos @ input fps, I get a video where e.g. with fps 60->15 The first 25% of the file play the whole video in 4x speed, and the last 75% are still image.

This PR would add a condition to only force output fps if `--ignore-frame-mismatch` is not set, since if you want to change output fps with FFmpeg you have to use this flag. I would also be down to add a flag to av1an to control fps forcing, or --- maybe even better --- add a flag to change fps directly through Av1an. This would additionally provide accurate progress bar data, which is broken with this method.

@Maintainers: please tell me which solution you would accept upstream, I will do my best to implement them :)

### DIFF
--- a/av1an-core/src/concat.rs
+++ b/av1an-core/src/concat.rs
@@ -162,7 +162,7 @@ pub fn mkvmerge(
     output: &Path,
     encoder: Encoder,
     num_chunks: usize,
-    output_fps: Rational64,
+    output_fps: Option<Rational64>,
 ) -> anyhow::Result<()> {
     // mkvmerge does not accept UNC paths on Windows
     #[cfg(windows)]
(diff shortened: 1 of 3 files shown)


t-nil:
fix formatting & rebase master

shssoichiro (approved):

PR #1024, opened 2025-06-23
Fix target-metric documentation inconsistencies

BoatsMcGee:
Documentation incorrectly suggests the `--target-metric` enum values are case insensitive. They must be in lowercase, which is consistent with other options.

butteraugli documentation and help output now mentions the target intensity is 203 nits for clarity.

Added missing XPSNR item from the list of metrics in Features->Target Quality documentation.


shssoichiro (approved):

shssoichiro:
It might not be bad if all options were made case insensitive, but that would be a separate discussion and changeset. The pain point for that change would also be that we have no control over whether that's the case for the encoders themselves, which may cause confusion to users.

so I'm good with this update.

Issue #994, opened 2025-05-19
Replace loadscript.vpy and VSPipe with vapoursynth-rs crate


BoatsMcGee:
I have an idea on how this can be implemented but am not familiar with all the components required so I'm looking for direct feedback on this.

Av1an already uses the vapoursynth-rs crate for retrieving video properties but does not maximize its utility for decoding or manipulating inputs. In PR #988, I implemented some functions in `vapoursynth.rs` that import an input video or custom VapourSynth script as well as the functions to manipulate the video for Scene Detection downscaling. This fixes the issue I discussed [here](https://github.com/rust-av/Av1an/pull/886#issuecomment-2810575054) where users that provide a custom VapourSynth script that does not contain the line Av1an is looking for to inject the downscaling code and the downscaling does not occur.

## Potential Implementation

Currently, Scene Detection uses av-scenechange `Decoder::VapourSynthDecoder` which takes a VapourSynth script as the input but creates and manages its own `Environment` and `Node`. In order to support manipulating the video, such as for downscaling, we need to instantiate our own `Environment` and import the node `Node` with the specified chunking-method, apply downscaling, and finally pass these instances to `Decoder::VapourSynthDecoder` so that we can use that decoder all in runtime without ever using vspipe. We can then expand this to also decode the video for Target Quality and during encoding.

This may require changes to the [rust-av/av-scene-change](https://github.com/rust-av/av-scenechange) repository and I'm not sure if it's within the scope of that project. If it isn't, then these changes may need to be implemented here instead.

## Other Issues

In another issue I mentioned that Target Quality and regular encoding sometimes fails when using multiple workers with a complicated VapourSynth script. I have seen 2 similar and possibly related issues: Loadscript.vpy is not found and the decoder (vspipe) exits and the script itself encounters an error related to the file system (accessing the cache file the chunking method uses). In the first case, VSPipe appears to be locking the loadscript.vpy file, causing another worker to fail when trying to access it. The same issue seems to occur with the second case, but within the script itself. Currently, I'm working around the issue by running only 1 worker at a time. Considering having multiple workers is the core feature of Av1an, this a likely critical issue that needs to be addressed.

Thanks,
\- Boats M.

PR #1030, opened 2025-06-26
Reduce Scene Detection loadscript Usage

### LINKED ISSUE #1029: Add ability to pass args to scene detection source script
Currently there's no way to pass args to scene detection script, but it can be very useful, especially when using DGDecNV. It would allow to use DGSource as a fast HW accelerated scaler for scene detection without jumping through hoops of manually setting scale values for when source script filename is `scene_detection.vpy`.
Also, when using `.vpy` script as a source, the `--sc-downscale-height` doesn't honor the source height if the specified downscale height is larger than source's.

BoatsMcGee:
Resolves #1029
Addresses #994 but does not completely resolve it. This change only applies to Scene Detection and does not remove the original `loadscript.vpy`, only the copied Scene Detection loadscript.

~~Requires rust-av/av-scenechange#182 and so the crate has been updated temporarily to use the fork/branch for that PR for demonstration purposes.~~ Merged.
Requires rust-av/av-decoders#1. Make sure to update `av1an-core/cargo.toml` to the latest `av-scenechange` and `av-decoders` before merging this PR.

Uses the underlying rust-av/av-decoders#1 changes to modify the output video node for downscaling programmatically rather than blindly modifying the user-provided input VapourSynth script. `Input` has been modified to store the necessary details to generate a loadscript text each time it is needed. For user-provided VapourSynth scripts, the path is used to instantiate `av-decoders::VapoursynthDecoder::from_file(Path)`, which will ensure the CWD for the script is the script directory itself. This does not apply for Target Quality probe measurement because using `environment.eval_file()` causes occasional File System issues due to frequently accessing the VapourSynth script file. During validation, if there is a discrepancy between the input script location and the CWD, we present a new warning recommending the script be in the same location as the CWD.

For non-VapourSynth inputs that use a VapourSynth chunking method such as FFMS2, the script can be generated and passed to `av-decoders::VapourSynthDecoder::from_script()` just like the user-provided script. Generating at the endpoint reduces the log pollution recently introduced in PR #988 when the `script_text` property was added. Note: The user-provided script is not generated/read from disk every time it is requested. Doing so would thrash the file system with frequent reads so we instead cache the value when initializing `Input::VapourSynth`. The aforementioned "log pollution" is not resolved for this particular input for this reason.

`register_node_modifier` takes a callback that lets us modify the output node programmatically and return it to use instead of the node evaluated from the just the script. This allows us to programmatically apply downscaling without resorting to copying and modifying a user-provided script with the risk of an invalid modification.

Current Scene Detection Decoding Scenarios:

1. User-provided VapourSynth script inputs with `--sc-downscale-height` specified: Apply downscaling using the most reliable method, `VapoursynthDecoder::from_file(Path)` and `register_node_modifier(callback)`
2. User-provided VapourSynth script inputs with no downscaling or Video inputs with a VapourSynth chunking method: Use VSPipe with provided script or loadscript as before
3. Y4M video remains the same: using FFmpeg to downscale when `filters` vector is not empty

### Additional Changes

`clip_info` caching requires us to retrieve the most accurate info the first time, so the input using a VapourSynth chunk method will be cached during validation when `clip_info` is first called. For some inputs, this could take a long time. I added an `info!()` log when this happens but was not able to see it. When DGDecNV is cached, we already had a log so the user has some feedback but the aforementioned `info!()` log does not show, despite initializing the logging system before validation. LSMASH and FFMS2 are fast enough that it may not be an issue, but BestSource indexing/caching is a lengthy process with no feedback.

The loadscript.vpy has been updated to ensure the downscale height is an int. The core thread limit has been removed for Scene Detection only - the limit still applies during encoding/probing. With a core thread limit Scene Detection decode speeds are negatively impacted.

### Caveats

Due to the slower decoding speed from `VapoursynthDecoder`, it is only used when we need it for the reliability. The following details only apply when the user provides their own VapourSynth script as input along with a Scene Detection downscale height.

Currently, `av-decoders` appears not to take advantage of the entire systems resources, leading to a slower than current Scene Detection performance. This could potentially be resolved in `av-decoders` and/or `av-scenechange`. However, the system resource usage with this PR appear to be significantly lower. Here is some testing and data below:

The following were using **`-m ffms2 --sc-downscale-height 720`** on `1920x1080 @ 23.976 fps, YUV420P10LE, SDR`.
Before (`vspipe.exe loadscript.vpy | av1an y4m`):
`00:00:30 ▌  10% 3253/34095 (107.77 fps, eta 5m)`
After (`av-scenechange` + `av-decoders::VapoursynthDecoder` + `register_node_modifier()`):
`00:00:30 ▌   5% 1856/34095 (60.87 fps, eta 9m)`

**`--sc-downscale-height 540`**
Before:
`00:00:30 ▌  12% 3993/34095 (131.61 fps, eta 4m)`
After:
`00:00:30 ▌   8% 2596/34095 (85.67 fps, eta 6m)`

I ran a similarly long test and monitored my resource usage in Task Manager as a quick sanity check on general performance.
Before: Total usage is ~80% of my entire machine, which is all of the available resources, presumably limited by CPU. Av1an uses ~10% CPU, vspipe uses 70%.
After: Total usage is ~40% of my entire machine, which is *not* all of the available resources, presumably limited by implementation. Av1an uses anywhere from 35%-50% but mostly around 40%.

Given how low this resource usage is, I believe Av1an could be significantly faster if we resolve the decoding speed in the underlying dependency. Av1an code should need no change for when that happens besides bumping the dependency version.

Thanks,
\- Boats M.


shssoichiro:
It seems like there may be some changes needed to resolve the CI issues. In #1061 you suggest waiting for this PR in order to merge that one, but since that one is small and resolves a user-facing bug, I'm going to go ahead and merge that one and leave a note here as well to remove that `None` from the clip_info call.

BoatsMcGee:
I just opened PR rust-av/av-scenechange#191 which should let us finally update the `av-scenechange` and `av-decoders` so the compile/tests shouldn't instantly fail.

I'm currently using this locally:

```toml
av-decoders = { version = "0.3.0", features = ["ffmpeg", "vapoursynth"] }
av-scenechange = { git = "https://github.com/BoatsMcGee/av-scenechange", branch = "bump-av-decoders", version = "0.17.1", default-features = false, features = [
    "ffmpeg",
    "vapoursynth",
] }
```

av1an-core/src/context.rs
```diff
@@ -82,9 +82,9 @@ impl Av1anContext {
         args.validate()?;
 
         let mut this = Self {
-            frames: args.input.clip_info(None)?.num_frames,
+            frames: args.input.clip_info()?.num_frames,
             vs_script: None,
-            vs_scd_script: None,
+            // vs_scd_script: None,
```
shssoichiro: Commented line that should be removed instead.

av1an-core/src/context.rs
```diff
@@ -181,47 +181,68 @@ impl Av1anContext {
...
+                    dec.set_variables(variables_map)?;
+                    av_scenechange::Decoder::from_decoder_impl(
+                        av_decoders::DecoderImpl::Vapoursynth(dec),
+                    )?
+                },
+                Input::Video {
+                    ..
+                } => av_scenechange::Decoder::from_script(
+                    &self.args.input.as_script_text(
+                        self.args.sc_downscale_height,
```
shssoichiro, line 232: Would building the script with sc_downscale in it result in issues with clip_info caching? i.e. if we are using both the same input and temp path for SC detection and the real encode.

BoatsMcGee: The `CacheKey` contains the Input instance but the `Input::Video` instance doesn't store the generated loadscript text so the cache should hit the same. The loadscript text is generated each time it is needed. Only the user-provided VapourSynth input script_text is cached in `Input::VapourSynth` but that value never changes so there should be no cache miss.

shssoichiro: My concern is more that, will the cache key end up being the same when we don't want it to be? For example, will the following scenario occur?

- Av1an generates load script for SC detection including downscaling
- This triggers the clip_info initialization and storage in cache
- Av1an then generates load script for encoding, which doesn't include downscaling
- Will this hit the same cache key? We want it to not hit because of the change in scaling.

BoatsMcGee: I see the potential issue now. The downscaling only occurs when the decoder is built with `Command.env("AV1AN_PERFORM_SCENE_DETECTION", "true")`. The info doesn't change as a result of this downscaling. It will initially generate assuming `AV1AN_PERFORM_SCENE_DETECTION` is false, so the video is directly output.

When accessing `clip_info`, we always assume no other operation has modified it and that applies for Scene Detection with downscaling as well. We have no code that accesses `clip_info` to retrieve information about the clip *as if it were downscaled*. We have no need for it either.

In short, when the initial `clip_info` is retrieved, it retrieves the info without any Scene Detection options/modifications.

PR #1037, opened 2025-06-30
feat: Set Number of Threads for VScore to 1 When No VPY File Is Specified

### LINKED ISSUE #1034: [Feature Request] Set Number of Threads for VScore to 1 When No VPY File Is Specified
vscore when threads is > 2 will allocate extra memory for parallel workers. When no vpy file is specified and the only thing that's done is reading the source video with ffms or best source, lowering the number of threads could decrease the amount of ram used dramatically per worker without hurting decoding performance.

A simple test to see this behavior is to make a vpy file that does nothing but reads a video file with either ffms or best source and change the number of threads in vscore from the default (that i am unaware of) to 1.

If the user specified the number of threads in their vpy file it should still however be respected as more threads is sometimes required for complex filter chains to run at acceptable speeds.

shssoichiro:
Closes #1034

### DIFF
--- a/av1an-core/src/loadscript.vpy
+++ b/av1an-core/src/loadscript.vpy
@@ -4,6 +4,8 @@ from vapoursynth import core
 
 # Set cache size to 1GB
 core.max_cache_size = 1024
+# Limit to one thread
+core.num_threads = 1
 
 source = os.environ.get('AV1AN_SOURCE', None)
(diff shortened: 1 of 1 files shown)


shssoichiro:
Looks like it has to be set both on the core and on the source filter. The source filters don't honor the `core.num_threads` setting.

PR #1041, opened 2025-06-30
Fix incorrect comment

shssoichiro:


shssoichiro:
ngl this is the kind of stuff I would expect Copilot to catch, but it didn't

PR #1048, opened 2025-07-02
replace TQ tolerance with configurable ranges

emrakyz:
Hammer the moldy chains that the developers shackled on the users. Put an end to the suffocating tyranny and give the users the long-awaited freedom.

Shatter the rigid constraints that imprisoned creative workflows in a cage of arbitrary limitations. Break free from the oppressive single-value targeting that forced users into narrow, inflexible quality corridors. Demolish the ancient walls of hardcoded tolerances that stood between encoders and their rightful control over quality ranges. Unleash the powerful flexibility that users have desperately craved, replacing the stale dictatorship of fixed thresholds with the sweet liberation of configurable boundaries.

- Replace fixed 1% tolerance with user-configurable target quality ranges
- Change `--target-quality` from single value to range format (`75-85` or `1.0-1.5`). Single values will continue working.
- Add `--qp-range` parameter to constrain quantizer search bounds independently. This makes it name more expressive/intuitive since `--min_q` can be mistaken with "quality". This will also make it similar to TQ range and it will help us remove a parameter/flag completely.
- Maintain backward compatibility by auto-generating range for single target values.

### DIFF
--- a/av1an-core/src/broker.rs
+++ b/av1an-core/src/broker.rs
@@ -221,9 +221,10 @@ impl Broker<'_> {
             update_mp_msg(
                 worker_id,
                 format!(
-                    "Targeting {metric} Quality: {target}",
+                    "Targeting {metric} Quality: {min}-{max}",
                     metric = tq.metric,
-                    target = tq.target
+                    min = tq.target.0,
+                    max = tq.target.1
                 ),
             );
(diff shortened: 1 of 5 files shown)


shssoichiro:
OUR target quality

image

shssoichiro (approved):

Issue #1062, opened 2025-07-11
Scene File Indentation Removed


Ironclad17:
When av1an receives a scene file input, it overwrites the file with the indentations removed which makes it difficult to troubleshoot or recheck individual scenes.

[Original scene file passed.json](https://github.com/user-attachments/files/21187415/Original.scene.file.passed.json)
[Scene file after encoding.json](https://github.com/user-attachments/files/21187416/Scene.file.after.encoding.json)

BoatsMcGee:
Closing as #1063 resolves this. Thanks!

PR #1063, opened 2025-07-12
Export Scenes as Pretty Json

NandeMD:
Solves #1062

### DIFF
--- a/av1an-core/src/context.rs
+++ b/av1an-core/src/context.rs
@@ -805,9 +805,9 @@ impl Av1anContext {
                 &self.vs_scd_script,
                 &zones,
             )?;
+            self.scene_factory.write_scenes_to_file(scene_file)?;
         }
         self.frames = self.scene_factory.get_frame_count();
-        self.scene_factory.write_scenes_to_file(scene_file)?;
         self.scene_factory.get_split_scenes()
     }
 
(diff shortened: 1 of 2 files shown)


BoatsMcGee:
Welcome! Thank you for getting this done - it looks good to me. I have just one addition and I think we can move it out of Draft, merge, and close this up.

Every time Av1an runs, it reads an existing scenes.json file or generates a new one. After scenes are read or generated, they are saved to the specified path. We only want this to happen when the scenes are initially generated so that the overwriting doesn't occur on subsequent executions of Av1an. We can do that with the following diff:

```diff
diff --git a/av1an-core/src/context.rs b/av1an-core/src/context.rs
index c4b5067..ee9fcb6 100644
--- a/av1an-core/src/context.rs
+++ b/av1an-core/src/context.rs
@@ -805,9 +805,9 @@ impl Av1anContext {
                 &self.vs_scd_script,
                 &zones,
             )?;
+            self.scene_factory.write_scenes_to_file(scene_file)?;
         }
         self.frames = self.scene_factory.get_frame_count();
-        self.scene_factory.write_scenes_to_file(scene_file)?;
         self.scene_factory.get_split_scenes()
     }
 

```

Please apply this patch ([scene-overwrite.patch](https://github.com/user-attachments/files/21199751/scene-overwrite.patch)) using `git apply .\scene-overwrite.patch` if you agree with the change.

Thanks,
\- Boats M.

NandeMD:
> Please apply this patch ([scene-overwrite.patch](https://github.com/user-attachments/files/21199751/scene-overwrite.patch)) using `git apply .\scene-overwrite.patch` if you agree with the change.

Hello and thanks! I applied the patch.

PR #1079, opened 2025-07-15
Add score-based enhanced extra splits method

### LINKED ISSUE #1035: [Feature Request] Split Video Scenes at Highest-Ranked Threshold by Scene Method Instead of Even Splits
Currently if a scene is longer than specified by extra splits in av1an, av1an will split the scenes till it is under the specified limit. This however can cause splits in awkward or illogical places like in the middle of a purely static scene or in a lot motion shot. This is more pronounced when encoding drawn content line anime that does not have constant moment or camera shake.

shssoichiro:
The goal of this change is to improve the locations chosen for extra splits, rather than being exactly in the middle, to being weighted based on a combination of scenecut cost and distance from the middle.

This fixes the distance calculation from the previous version.

Closes #1035

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,8 +9,11 @@ once_cell = "1.8.0"
 path_abs = "0.5.1"
 tracing = "0.1"
 
-[profile.dev.package.av-scenechange]
-opt-level = 3
+[profile.dev.package."*"]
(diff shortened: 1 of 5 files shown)


shssoichiro:
Okay attempt 2 let's go

PR #1089, opened 2025-07-17
Fix: Absolut Path generation on Linux when using dgdecnv

Tyr3al:
Hi,

this small PR fixes an issue, that occurs when using dgdecnv on a linux system (Arch Linux 6.15.5-arch1-1 in my case). When starting av1an like this `av1an -i test.mkv -o output.mkv -m dgdecnv`, av1an crashes with a file not found error:

> Error: No such file or directory (os error 2)

After checking av1an with strace -f, I could see that av1an was trying to access the temp directory at a point in time, when it was not created yet.

Stepping through the code, I could pin point it to the  `to_absolute_path` method, that uses `path.canonicalize()` under linux. This method does not seem to work on files (or directories) that are non-existant. Unfortunately, `to_absolut_path` is called in `vapoursynth.rs` for dgdecnv before the temp directories have been created. 

Instead of changing the order of initialization i just removed the inline function and changed the code to use `absolute(path)` for all operating systems. This function seems to work just fine under linux.

Checklist:

- [x] Run cargo check
- [x] Run cargo clippy
- [X] Run cargo +nightly fmt
- [x] Run cargo build
- [x] Run cargo test

### DIFF
--- a/av1an-core/src/util.rs
+++ b/av1an-core/src/util.rs
@@ -1,9 +1,7 @@
 #[cfg(test)]
 mod tests;
 
-use std::{
-    path::{ Path, PathBuf},
-};
+use std::path::{Path, PathBuf};
 
 /// Count the number of elements passed to this macro.
 ///
--- a/av1an-core/src/scene_detect.rs
+++ b/av1an-core/src/scene_detect.rs
@@ -1,10 +1,11 @@
 use std::{
     collections::BTreeMap,
     io::{IsTerminal, Read},
+    path::absolute,
     process::{Command, Stdio},
     thread,
 };
-use std::path::absolute;
+
 use anyhow::bail;
 use av_decoders::{DecoderError, DecoderImpl, VapoursynthDecoder, Y4mDecoder};
 use av_scenechange::{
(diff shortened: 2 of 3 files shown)


shssoichiro:
Thanks! You'll need to run cargo fmt with the nightly compiler. If you're using rustup you can do this using `cargo +nightly fmt`

Tyr3al:
@shssoichiro Thanks for the hint. I've run fmt in nightly mode and pushed the result.

av1an-core/src/util.rs
```diff
@@ -139,12 +136,3 @@ pub fn read_in_dir(path: &Path) -> anyhow::Result<impl Iterator<Item = PathBuf>>
         })
     }))
 }
-
-#[inline]
-pub(crate) fn to_absolute_path(path: &Path) -> io::Result<PathBuf> {
-    if cfg!(target_os = "windows") {
-        absolute(path)
-    } else {
-        path.canonicalize()
```
shssoichiro, line 148: I don't know why we were doing this... tbh the Rust docs should just have a big warning on the `canonicalize` function that says DO NOT USE THIS 😆

PR #1105, opened 2025-08-02
Bugfix: Exit target quality if bounds are meet

Thomasedv:
From Discord discussion. 

Av1an can underflow the min q values when a lot of probes are used, and the target metric fails to reach the desired quality even at low.

### DIFF
--- a/av1an-core/src/target_quality.rs
+++ b/av1an-core/src/target_quality.rs
@@ -256,7 +256,7 @@ impl TargetQuality {
             }
 
             // Ensure quantizer limits are valid
-            if lower_quantizer_limit > upper_quantizer_limit {
+            if lower_quantizer_limit >= upper_quantizer_limit {
                 log_probes(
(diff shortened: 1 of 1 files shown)


emrakyz:
Your PR is wrong.
This doesn't fix the actual underflow issue. It only changes when to exit the loop. The underflow still happens when computing next_quantizer - 1 when next_quantizer is 0.

- Start with lower_quantizer_limit = 0, upper_quantizer_limit = 70
- If scores keep being too low (below target range), the upper limit keeps decreasing
- Eventually, next_quantizer becomes 0
- When score is still below target, it tries 0 - 1 which underflows
- This causes the subsequent predict_quantizer calls to return huge values

Your PR doesn't fix it:
- The arithmetic underflow still occurs before the bounds check
- When lower = upper, it's still valid to probe that exact value, exiting is not a good idea
- The problem is the unchecked subtraction, not the loop exit condition

PR #1108, opened 2025-08-06
feat: add quarter-step CRF support & cleanups

emrakyz:
- Quarter-step CRF will be supported by `x264`, `x265` and `svt-av1`
- Especially `x26*` encoders will heavily benefit from the additional accuracy / speed
- Runtime detection of SVT-AV1 quarter-step support via test encoding
- Cache quarter-step capability check using `OnceLock`
- Maintain backward compatibility with mainline SVT-AV1
- Remove obsolete SVT-AV1 version check and legacy parsing


BoatsMcGee:
I see no reason to remove the already existing support for older SVT versions - even if most people aren't using it any longer. I would understand dropping support if there was never any checking to begin with but in this case the work was already done to keep supporting it and maintaining that functionality is as cheap as it gets. Furthermore, we have already discussed several times that this same functionality needs to be expanded for other versions of SVT where the highest Preset changed or the usage of `-lp` changed in SVT 3.0.0. Both of these are known issues we planned to address eventually. Addressing them uses the functions removed in this PR

Some snippets from me discussing expansion of these functions:
> Well, that's unfortunate. We could handle this in 2 ways:
> * Change the `const MAXIMUM_SPEED_SVT_AV1: u8 = 12;` to `const MAXIMUM_SPEED_SVT_AV1: u8 = 10;` - this is lazy and would introduce unexpected behavior for users who are still using older encoder builds
> * Add a `const MAXIMUM_SPEED_SVT_AV1_3_1: u8 = 10;` and handle accordingly after parsing the version

> ... I did notice this was an issue but since SVT clamps the value silently it wasn't that much of a problem.
> We should make a match arm specifically for ^3.X.X when the `--lp` changes were introduced. ...

If we decide to start working on the above issues then we'll have to restore that code regardless.

It also looks like you accidentally committed some log files so you may need to add those paths to the `.gitignore` and remove them from the PR.

av1an-core/src/encoder/mod.rs
```diff
@@ -732,89 +711,23 @@ impl Encoder {
                 "--kf-max-dist=9999"
             ],
             Self::svt_av1 => {
-                if *USE_OLD_SVT_AV1 {
```
BoatsMcGee: I don't see a technical reason to remove this feature.

shenef:
SVT-AV1 3.1 brought back presets 11 and 12 in `--rtc` mode:
https://gitlab.com/AOMediaCodec/SVT-AV1/-/merge_requests/2452

av1an-core/src/encoder/mod.rs
emrakyz: Not technical; yes. It's cleanup. I think we even need to remove more code. They can be added when needed.

"old" svt-av1 mentioned here is before v0.9.0 meaning years ago.

I really think the support should be stopped. Even huge corps drop support for this "old".

A better idea would be to remove all version / max preset types of checks and functionality because they add no functionality to av1an itself. It's a script. We need to let the encoder handle its own validation.

If something is not as universal (such as a default parameter that is not included in other versions); just remove that default parameter since it's not universal.

My idea in mind is to even remove the quarter step checking after mainline adds it in a release version. 

It's simply a "works always, fails fast principle".

So my idea is:
- av1an orchestrates encoding workflows, encoders handle their own parameter validation
- If there's an incompatibility, encoders fail immediately with clear error messages
- Don't try to be clever about version differences
- Only use default parameters/features that work
- When encoders change, users update encoders, not av1an

Maintainability:
- No complex version detection logic that becomes stale
- No conditional code paths based on encoder versions
- Easier to reason about and debug
- Fewer edge cases and corner cases to handle
- More predictable behavior
- Less surface area for bugs
- Clear, immediate feedback when something is wrong
- No "it works on my machine" issues due to version-specific workarounds

BoatsMcGee: Av1an has validation for specific encoders outside of what's been removed here so claiming validation is not a part of Av1an or its responsibilities is inaccurate. It's used for the benefit of the user and in ensuring reliability. The validation isn't unnecessary either, it prevents Av1an from using the wrong parameters because it wouldn't have knowledge of the encoder version. Presumably, the older SVT parameters were either necessary or at least highly recommended and that is why that code exists. Some parameters used in newer SVT versions may not be accepted by the older version so we need some knowledge of the encoder version in order to know which set of parameters to use. It's very likely we simply cannot assume that 1 set of parameters works across all versions which is why this code was added in the first place. There's no good reason to remove that code - it's at best trading developer convenience for more runtime debugging. This code isn't inconvenient anyway, nor does it become stale - it's just a very simple version parser. Versions don't change their behavior so the logic relying on semver information can't become stale. Removing that information doesn't help with debugging either unless the user is willing to provide their encoder information in addition to their debug logs - which from my experience so far we rarely get unprompted. Rather, that version information should be presented in the `--version` and in the debug logs to help with debugging because there is no guarantee that the encoder error contains the version if it is even captured in the first place. Removing that information doesn't make it easier to debug either because if it was behaving unexpectedly and we have no information about the version then it would be more guesswork - not less.

Dropping support for an older encoder version would make sense if there were security concerns or keeping it prevents adding functionality or the work to support it isn't worth the effort. That isn't that case here as there are no security concerns for older SVT versions that I'm aware of, this PR isn't affected by the existence of the parser, and the effort was already made so removing it just erases that functionality with no improvement or replacement. If the feature was a work in progress then stopping development and removing the incomplete feature would make sense but that's not the case here either. And the expansion of that exact code is necessary to address previously reported issues so this is actually removing what is planned to be the basis for future fixes and features.

If we move all responsibility of encoder validation to the encoder itself and the end-user, then you could also argue that there is no need to make any distinction in which encoder is being used as the user would be responsible for ensuring their `--video-params` are appropriate for their `--encoder-binary SvtAv1EncApp`. However, that's not the case for Av1an as it does need and use that information in its logic. Until just recently, the CPU-Used/Speed/Preset value for Target Quality probes was set based on this logic. If that logic were removed, the value would be inaccurate. I have discussed plans to return that functionality for a different purpose so removing the version information would make that feature impossible along with the other fixes and features mentioned.

> When encoders change, users update encoders, not av1an

This would make sense if Av1an bundled encoders, did not use the System binaries, did not allow custom binaries, and specified that Av1an only works with specific versions. However, that's not the case here either and I would wager many users are not using the absolute latest binaries - not even the recent 3.x.x SVT binaries which introduced breaking API changes that require Av1an to use this exact function to provide good input.

There is no improvement to feedback when this information is removed. That reason doesn't apply to this situation. Nor does "it works on my machine" because the lack of version-specific parameters means it works for fewer environments than it did before. Before, we had support for older SVT and now we don't. That is not an improvement. That is a regression.

You cannot claim this change makes Av1an always work because the logic removed was originally there to increase the chances that Av1an will work by supporting 2 different APIs. 1 size does not fit all which is why we need this functionality. Additionally, you cannot call a feature removal "cleanup" because "cleanup" is typically used for removing unused/dead code.

emrakyz: I think both of our reasons are valid.

But the code can be adapted to whatever method is used. I really think even removing all encoder specific behavior is possible and they could be applied differently without requiring hard checks or breaking compatibility. But of course, it's not the goal of this PR.

I can for now add them back.

What I said shouldn't have been taken as is though. What I propose was something gradual: Remove everything in av1an that makes decisions based on a specific version. Things can be handled in easier ways.

We now have zones, overrides, --probing-video-params, and video params. ALL should be handled by the user completely freely. av1an shouldn't do anything by its own.

Well, I even dislike this PR of my own since it requires checking for quarter steps. This also could have been handled differently (even with an alternative param or an encoder path) but current av1an can make things more complex then it should be.

emrakyz:
@shenef 

Yes and it's out of context for av1an and I reverted it.

av1an-core/src/encoder/mod.rs
BoatsMcGee: Please move the feature removals into a separate PR so they can be evaluated in isolation. I haven't tested anything but the rest of PR looks fine and shouldn't be held up because you decided to bundle multiple changes into a single PR. 

Av1an cannot perform some of its functions without knowing the encoder version. Even more so for future planned functionality/fixes. I see no alternative for this particular code. The existing function was already as simple as it gets and the value is cached in a static variable so there's no performance concerns about running the validation multiple times each time the information is requested. Please offer your better alternative in that PR instead so we can evaluate it.

I don't understand; If you're adamant that there's a better way to do it, then why didn't you demonstrate that with a better replacement in this PR instead of just outright removing it? There's no need to rush these changes - especially when you are already aware that we're currently focusing on bugfixes and QoL improvements in preparation for the next release.

> Zones, overrides, video params

The version check doesn't affect users who are using those features. It's for the users who *don't* provide them because `--video-params` isn't a required parameter, it's an option. Av1an needs to know what the appropriate defaults are for a given version in order to use them without encountering an error. Av1an also doesn't explicitly tell users they can modify the Scenes JSON `.zone_overrides` value and it doesn't validate those values anyway so there's no concern for Av1an over-validating things for that.

If you're advocating for removing all Av1an validation then the program would become less reliable and less convenient. I just added necessary validation for Zones for added Target Quality parameters. Removing that validation would mean that the user would run Scene Detection and start encoding before encountering an error that Av1an would have caught beforehand and prevented from occurring in the first place. Imagine you're that user who is trying to debug why it's crashing when it's encoding, you would have to somehow realize that VMAF wasn't enabled for your FFmpeg based on reading the debug log - assuming you know how to and are remotely interested in doing so. There's also no guarantee that FFmpeg would provide a useful output or error message that makes it obvious. Assuming so would be naive.

Again, I have to recommend making multiple PRs and limiting scope. It's fine if you want to perform cleanups and refactors but you shouldn't be forcing them at the same time as you're adding features. Merging PRs are an all or nothing affair, maintainers can't cherry-pick changes.

shssoichiro:
Looks fine to me, let's see what useless comments Copilot has

av1an-core/src/encoder/mod.rs
```diff
@@ -86,35 +125,40 @@ pub(crate) fn parse_svt_av1_unprocessed_tokens(output: &[u8]) -> Option<Vec<Stri
...
-        let out = if output.stdout.is_empty() {
-            output.stderr
-        } else {
-            output.stdout
-        };
-        // assume an old version of SVT-AV1 if the version and unprocessed tokens failed
-        // to parse, as the format for v0.9.0+ should be the same
-        parse_svt_av1_unprocessed_tokens(&out).is_none_or(|tokens| tokens.is_empty())
+            Command::new("SvtAv1EncApp")
+                .args(["-i", test_file.to_str()?, "--crf", "50.75", "-b", NULL])
```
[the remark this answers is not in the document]
shssoichiro: Copilot... you are stupid.

av1an-core/src/target_quality.rs
```diff
@@ -158,28 +158,34 @@ impl TargetQuality {
...
                     _ => target,
                 },
                 self.interp_method,
+                step,
             )?;
 
             if let Some((quantizer, score)) = quantizer_score_history
                 .iter()
-                .find(|(quantizer, _)| *quantizer == next_quantizer)
+                .find(|(quantizer, _)| *quantizer == next_quantizer as f32)
```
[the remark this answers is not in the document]
shssoichiro: These might actually be true since we changed the return type of `predict_quantizer`. @emrakyz can you verify? Surprised clippy did not trigger on these though.

Issue #889, opened 2024-11-04
Feature Request: Add keepawake-rs to prevent system suspend during encodes


Bl4cKn1gh7:
I've been using [keepawake](https://github.com/segevfiner/keepawake-rs) on windows during encodes to prevent my system from sleeping during encodes. Its cross platform and there's a [rust crate](https://docs.rs/keepawake/latest/keepawake/) available, so it might be a solid addition for convenience.

I use the idle option so the display still turns off which is reasonable, though apparently the difference between idle and sleep varies depending on OS according to upstream.

PR #1109, opened 2025-08-09
Sleep Guard

Tyr3al:
I had the problem, that av1an encodes would not run through overnight, as my system would simply go into sleep - even though an encoding job was already running. In my case that would be Arch Linux. It seems like other people have this issue as well, see #889.

This PR ads a sleep guard to av1an that should prevent sleeping while an encode is running. I first tried to use the suggested crate [keepawake](https://docs.rs/keepawake/latest/keepawake/) as suggested in the Issue, but it pulled far more dependencies as expected, which were also outdated. 

I sat down to write implementations for linux, windows and macOS myself, that should prevent the system from going into sleep.

I have tested the linux and the windows implementations. The macOS implementation is untested, as I was not able to get av1an to start an encoding job there. 

Screenshot Linux (KDE Plasma):
image

Screenshot Windows 10:
Unbenannt

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,12 +256,15 @@ dependencies = [
  "av1-grain",
  "cfg-if",
  "colored",
+ "core-foundation",
  "crossbeam-channel",
  "crossbeam-utils",
  "ctrlc",
  "dashmap",
+ "dbus",
  "indicatif",
  "itertools",
+ "libc",
  "memchr",
  "nom 8.0.0",
  "num-traits",
@@ -284,6 +287,7 @@ dependencies = [
  "tracing",
  "vapoursynth",
  "which",
+ "windows",
  "y4m",
 ]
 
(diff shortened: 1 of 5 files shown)


emrakyz:
Thanks for your PR!

This seems important but can't you handle it on the OS level instead with much less complexity?

And screen locking would be a needed functionality while it runs.

Tyr3al:
> This seems important but can't you handle it on the OS level instead with much less complexity?
> 
AFAIK, there are some things you can do outside of the application to prevent sleep. This PR is rather a convenience function, so that you do not have to call av1an from another wrapper. E.g. in linux:

`systemd-inhibit --why="Prevent sleep during execution" --what=sleep ./av1an`

> And screen locking would be a needed functionality while it runs.

The screenshot may be misleading. The sleep inhibitors I'm creating only prevent sleep. The system is still able to turn off and lock the screens.

PR #1120, opened 2025-08-17
Change default encoder from aomenc to svt-av1

### LINKED ISSUE #1057: Change the default encoder from aomenc to SVT-AV1
Per my understanding, it's generally considered to be better and faster than aomenc.
Recent updates to av1an also have been largely SVT-AV1 focused and the aomenc implementation currently also has some outdated defaults and is missing proper support for Target Quality which doesn't make it a good candidate for being the default encoder.

shenef:
Closes #1057


shssoichiro (approved):
Thanks!

PR #1165, opened 2025-11-28
Support new svt-av1 progress format

shssoichiro:
svt-av1 introduced a new progress format in https://gitlab.com/AOMediaCodec/SVT-AV1/-/merge_requests/2511. This change supports parsing that format, and supports fallback parsing for the old format.


shssoichiro:
Copilot reviews are paid only now? I guess giving it to us for free was just a limited time jebait. Great work, Microsoft.

We gotta move to Codeberg.

PR #1178, opened 2025-12-15
Update dependencies and fix clippy lints

shssoichiro:
also remove deprecated authors field from `Cargo.toml` since crates.io already knows the crate owners


shssoichiro:
I have no idea why there are failures in CI that I can't produce locally

shssoichiro:
Oh no, I think it's failing now because the docker image auto-updated to vapoursynth r73. Well... I guess I need to prioritize fixing that.

PR #1181, opened 2025-12-24
Fix 8-bit encode when probing rate is 1

Uranite:
When the target quality feature is used, and the probing rate is 1, the encode (and the probing) is 8bit, even though Av1an's default is 10bit (yuv420p10le)
This PR fixes this issue by running compose_ffmpeg_pipe regardless of the probe rate
Now the encode is 10bit when the target quality feature is used even when the probing rate is 1

ffmpeg adds about 200MB of memory per worker
Ideally we should be using VapourSynth for this for less overhead


shssoichiro:
Thanks, seems good to me though I need to fix the CI first (breakage is certainly related to Vapoursynth)

PR #1184, opened 2025-12-31
Parse `nb_frames` from ffprobe as string

### LINKED ISSUE #1124: Error: Failed to get number of frames for input video
Hey!

i am trying to get this to run in windows.
i have the av1an.exe from 3 days ago and a newly built aomenc.exe

when trying to run the command:

`av1an -i "original.mp4" -o "av1an_test.mp4" --encoder aom --target-metric vmaf --target-quality 96`

i get an error telling me:
Error: Failed to get number of frames for input video
Caused by:
    invalid type: string "144", expected usize at line 15 column 30

seems like something inside the frame counting gives back a string instead of an int?
can i do something about this, or do i need to wait for another relase? :)

thanks!
Mem

Wallunen:
This bug makes it impossible to use an MP4 file as input, but MKV files and WebM files work as expected.

Closes #1124

### DIFF
--- a/av1an-core/src/ffmpeg.rs
+++ b/av1an-core/src/ffmpeg.rs
@@ -49,7 +49,7 @@ struct FfProbeStreamInfo {
     pub pix_fmt:        String,
     pub color_transfer: Option<String>,
     pub avg_frame_rate: String,
-    pub nb_frames:      Option<usize>,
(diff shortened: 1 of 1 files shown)


shssoichiro:
Thanks, this makes sense now that I see what is happening.

PR #1185, opened 2026-01-02
Version 0.5.2

shssoichiro:


Wallunen:
Are tag names gonna be consistent from now on? The "v" prefix makes automated update checks easier.

shssoichiro:
That is the intent, yes. I didn't even realize we had been doing it without the `v` prefix until recently.

PR #1188, opened 2026-01-08
create scene detection folder if not exist

Khaoklong51:
Scene detection will failed with `Error: The system cannot find the path specified. (os error 3)` at write file if folder that point to scene detection file does not exist. This pr ensure folder is create before program write scene detection file

### DIFF
--- a/av1an-core/src/scenes/mod.rs
+++ b/av1an-core/src/scenes/mod.rs
@@ -3,7 +3,7 @@ mod tests;
 
 use std::{
     collections::{BTreeMap, HashMap},
-    fs::{create_dir_all, File},
+    fs::File,
     io::Write,
     path::Path,
     process::{exit, Command},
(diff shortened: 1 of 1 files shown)


av1an-core/src/scenes/mod.rs
```diff
@@ -426,6 +426,14 @@ impl SceneFactory {
             bail!("compute_scenes must be called first");
         }
 
+        if let Some(parent) = scene_path.as_ref().parent() {
+            if !parent.exists() {
+                if let Err(e) = create_dir_all(parent) {
+                    bail!("Cannot create folder at {}: {}", parent.display(), e);
+                }
```
shssoichiro: It seems we have a `create_dir!` macro in av1an that we can use here instead, which includes logging on error.


```suggestion
                create_dir!(parent)?;
```

PR #1193, opened 2026-01-17
Add a Windows application manifest

Khaoklong51:
Add a Windows application manifest using [embed-resource](https://crates.io/crates/embed-resource). 

This enable long path support on windows without manifest av1an failed with the following error.
```
Error: Failed to get number of frames for input video

Caused by:
    0: get_clip_info
    1: An error occurred in VSScript
    2: Python exception: Source path does not exist

       Traceback (most recent call last):
         File "src/cython/vapoursynth.pyx", line 3393, in vapoursynth._vpy_evaluate
         File "src/cython/vapoursynth.pyx", line 3394, in vapoursynth._vpy_evaluate
         File "<string>", line 26, in <module>
       ValueError: Source path does not exist
```

### DIFF
--- a/av1an/Cargo.toml
+++ b/av1an/Cargo.toml
@@ -31,7 +31,7 @@ tracing-appender = "0.2"
 tracing-subscriber = { version = "=0.3.19", features = ["env-filter"] }
 
 [target.'cfg(windows)'.build-dependencies]
-embed-resource = "=3.0.6"
+embed-resource = "3.0.6"
 
 [build-dependencies]
 vergen-git2 = { version = "1.0.0", features = ["build", "rustc", "cargo"] }
--- a/av1an/build.rs
+++ b/av1an/build.rs
@@ -15,7 +15,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     #[cfg(windows)]
     embed_resource::compile("av1an.exe.rc", embed_resource::NONE)
-        .manifest_optional()
+        .manifest_required()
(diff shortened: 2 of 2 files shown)


av1an/build.rs
```diff
@@ -12,5 +12,11 @@ fn main() -> Result<(), Box<dyn Error>> {
         .add_instructions(&cargo)?
         .add_instructions(&rustc)?
         .emit()?;
+
+    #[cfg(windows)]
+    embed_resource::compile("av1an.exe.rc", embed_resource::NONE)
+        .manifest_optional()
```
shssoichiro: It is probably better to use `manifest_required` so that we don't silently exclude long path support if there's an issue with the resource.

av1an/Cargo.toml
```diff
@@ -30,6 +30,9 @@ tracing-appender = "0.2"
 # FIXME: Locked due to https://github.com/tokio-rs/tracing/issues/3369
 tracing-subscriber = { version = "=0.3.19", features = ["env-filter"] }
 
+[target.'cfg(windows)'.build-dependencies]
+embed-resource = "=3.0.6"
```
shssoichiro: We should prefer not to lock the exact version (remove the `=`). We are only doing this on `tracing-subscriber`  due to a bug in newer versions of that library that hasn't been fixed yet.

shssoichiro:
Thanks, looks good with a couple of minor suggestions

PR #1194, opened 2026-01-17
loadscript cleanup

Khaoklong51:
cleanup code in generating loadscript and loadscript file itself.


shssoichiro:
Thanks!
                                                                                                                             
Minor Observation (Non-blocking)                                                                                           
                                                                                                                             
The struct `LoadscriptArgs` at `vapoursynth.rs:717-726` still contains the fields:                                             
  - `scene_detection_downscale_height`                                                                                         
  - `scene_detection_pixel_format`                                                                                            
  - `scene_detection_scaler`                                                                                                   
                                                                                                                             
These fields are now unused within `generate_loadscript_text()`. While not a bug, they represent leftover code that could be cleaned up in a follow-up PR.

shssoichiro:
I can create the follow-up with the extra cleanup.

PR #1209, opened 2026-02-05
Disable mkvmerge chunking on Windows

Uranite:
We don't need mkvmerge chunking on Windows because we're using the json file
And up the chunks on Linux from 100 to 960
And disable intermediate file creation if there is only 1 chunk


shssoichiro:
Thanks!

PR #1255, opened 2026-07-05
Add VapourSynth extra plugin path to Dockerfile to fix FFMS2 loading

Eauldane:
With VapourSynth R74, the [packaging system has changed](https://www.vapoursynth.com/2026/03/26/new-packaging-and-install-methods-in-r74/). This means that plugins like FFMS2 aren't counted as loaded anymore - Arch still installs them at the same place, VapourSynth just needs an extra pointer to see them. This is just a quick fix that sets an ENV var during build that should solve that.

Tested and validated in a local transcode.


shssoichiro (approved):

shssoichiro:
Thanks!

PR #1260, opened 2026-08-07
Add Fritsch–Carlson and generalized pchip from xav; use pchip for 6th+ probes

Uranite:
and Validate 5th-round interpolation method

Code is taken from xav, and pchip for 6th+ probes is what xav does
Fritsch-Carlson test is the exact same as natural cubic spline
From my testing, with very tight target quality, like 79.999-80.001, probes are greatly reduced

### DIFF
--- a/av1an-core/src/interpol.rs
+++ b/av1an-core/src/interpol.rs
@@ -115,87 +115,121 @@ pub fn natural_cubic_spline(x: &[f64], y: &[f64], xi: f64) -> Option<f64> {
     Some(b_coeff.mul_add(dx, a_coeff) + c_coeff.mul_add(dx.powi(2), d_coeff * dx.powi(3)))
 }
 
-pub fn pchip_interpolate(x: &[f64; 4], y: &[f64; 4], xi: f64) -> Option<f64> {
-    // Check strictly increasing
-    for i in 0..3 {
+pub fn fritsch_carlson(x: &[f64], y: &[f64], xi: f64) -> Option<f64> {
+    let n = x.len();
+    if n != 3 || n != y.len() || xi < x[0] || xi > x[n - 1] {
+        return None;
+    }
+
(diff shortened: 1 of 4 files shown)


av1an-core/src/interpol.rs
```diff
@@ -115,87 +115,117 @@ pub fn natural_cubic_spline(x: &[f64], y: &[f64], xi: f64) -> Option<f64> {
...
-    for i in 0..3 {
+pub fn fritsch_carlson(x: &[f64], y: &[f64], xi: f64) -> Option<f64> {
+    let n = x.len();
+    if n != 3 || n != y.len() || xi < x[0] || xi > x[n - 1] {
+        return None;
+    }
+
+    let k = usize::from(xi >= x[1] && xi <= x[2]);
+
+    let d0 = (y[1] - y[0]) / (x[1] - x[0]);
```
shssoichiro, line 130: - **Problem:** `fritsch_carlson` does not verify that adjacent `x` values are strictly increasing. Probe history is sorted by score, but scores can be equal; for inputs such as `[80.0, 80.0, 80.0]`, the selected interval has zero width and `t` becomes `NaN`. Because the function returns `Some(NaN)`, `predict_quantizer` does not invoke its binary-search fallback. The added non-increasing test misses this because its `xi` fails the earlier bounds check.
- **Suggestion:** Reject non-increasing inputs before calculating slopes, matching the other interpolation functions:
  ```rust
  if x.windows(2).any(|pair| pair[1] <= pair[0]) {
      return None;
  }
  ```
  Add a duplicate-score regression case whose `xi` remains within the endpoint bounds.

Uranite: Fixed and added new tests

shssoichiro (approved):

shssoichiro:
Thanks!

PR #1261, opened 2026-08-12
fix: replace removed -vsync 0 with -fps_mode passthrough

p-linnane:
FFmpeg 9 removed `-vsync`, so Av1an fails with `Unrecognized option 'vsync'`. This swaps in its exact successor, `-fps_mode passthrough`, in the target-quality probe pipe and the stream-copy segmentation path, plus a regression test.

Note this raises the effective FFmpeg floor to 5.1, where `-fps_mode` was added. Since `-vsync` is gone entirely, no single flag spans that boundary.


p-linnane:
@shssoichiro In Homebrew we only support the latest versions of dependencies, so raising the effective FFmpeg floor doesn't affect our packaging. However, I would appreciate your thoughts on if this is acceptable upstream.

shssoichiro:
I'm definitely okay with deprecating versions older than 5.1. Thanks!

shssoichiro (approved):

